PR #38, opened 2014-12-22
plot: use httpfs to fetch table.txt and pipe to gnuplot

syockit:
Works so far. Probably needs some polishing as most of the returned errors are simply discarded.


barnex:
Thanks, merged.

PR #46, opened 2015-06-17
WIP energy minimizer

syockit:
Based on Exl et al., JAP 115, 17D118 (2014) doi:[10.1063/1.4862839](http://dx.doi.org/10.1063/1.4862839), fast energy minimizer suitably for hysteresis simulations.

The code is not yet ready for merge. I have some reservations about the implementation: 
1. The minimizer implements the Stepper interface, but is not manually accessible by the user. It could confuse the user as some of the variables are not used in the same way as the usual dynamic LLG steppers e.g. the time step unit is not physical. On the other hand, it allows for realtime visualization in the HTML viewer without extra modifications.
2. Magnetization is not normalized after each step. Although the method guarantees constant magnetization, numerical artifacts may build up.
3. The paper mentioned above only involves plain LLG. Current code adds in STT, which I'm not familar with. Thermal excitation is disabled (through flag `relaxing = true`).

Convergence checking is via magnitude of magnetization change per step (not time step).
The code is modular, although I had to make my own kernel since it involves division operation for which there are no suitable kernels.

It is still missing a test unit, but otherwise the code works fine as far as I have checked.


barnex:
I have already created a branch "minimizer" and merged your request there. Unit test added, the result is correct and it runs blazing-fast. Well done!

barnex:
As you may have noticed, I have cleaned up and documented the code a bit, merged into master and tagged version v3.9c.

I have tried a few hysteresis loops, as well as several unit tests and not only does it work well until now, it's really blazing fast. Great job!

I would suggest to send a little announcement to the mumax mailing list, as many users may be very happy with this feature.

syockit:
Thanks for the merge! I've also announced on the mailing list as well.

Sad to see you go though…

PR #48, opened 2015-10-17
Add CropRegion

syockit:
Wrapper for Crop that takes the smallest (x,y,z)-range that includes the specified region.

Also, naming for crop is modified, where previously when there are at least two ranges, there would be two underscores between them e.g. xrange33-55__yrange3-17. Now it is fixed to xrange33-55_yrange3-17.


barnex:
Looks good, thanks!

PR #57, opened 2016-02-09
mumax server to ignore files that start with "._"

jsampaio:


godsic:
Perhaps make it more general, i.e. Ignore hidden files in os-agnostic way?

jsampaio:
That could be all files that start by "." in linux (& mac). I didn't know how to do it, but I searched a bit and the current version ignores all files that start by a '.'. 
On windows, we'd have to check for the 'hidden' flag, and I can't find out how... maybe someone more expert on Go knows?

PR #82, opened 2016-10-28
let mumax3-convert convert wildcards if needed

JeroenMulkers:
Let Mumax3-convert expand wildcards which are not expanded by the shell.
(pointing a finger at cmd.exe)

### DIFF
--- a/cmd/mumax3-convert/main.go
+++ b/cmd/mumax3-convert/main.go
@@ -39,6 +39,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


godsic:
Have you tested on both Linux and Windows?

JeroenMulkers:
I do not have a Windows machine, but I cross compiled mumax3-convert for Windows and tested in Wine.

I also tested the modification on linux. Following commands, give the same output:
```
mumax3-convert -png *.ovf
mumax3-convert -png "*.ovf"
```

xgstation:
I have tested and it works.
Another comment: is it possible to use cuda to do the convert?

JeroenMulkers:
The gpu can't read or write files. So I doubt that using cuda in mumax3-convert will speed up the conversion much, if any at all.

PR #87, opened 2017-01-04
Bib

JLeliaert:
Add library file "references.bib", based on users' simulations, to the output directory.


godsic:
Whats this one is about?

JeroenMulkers:
Mumax3 returns a bibtex file with relevant references. This includes the paper about Mumax3. Depending on the used features in a simulation, other relevant papers will be added.

If you think that some references are missing, please add them to the library.

PR #91, opened 2017-01-24
Add several functions from the CUDA runtimeapi

jcvernaleo:
We've been using these extra functions from the runtimeapi in our cryptocurrency mining software https://github.com/decred/gominer for months now.  Figured they might be of some use for the upstream mumax libs.

### DIFF
--- a/cuda/cu/cgoflags.go
+++ b/cuda/cu/cgoflags.go
@@ -2,7 +2,7 @@ package cu
 
 // This file provides CGO flags to find CUDA libraries and headers.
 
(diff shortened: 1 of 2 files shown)


barnex:
Cool. Could you please re-address this request to https://github.com/barnex/cuda5, which is the same code, but maintained outside of mumax. The cuda bindings in mumax are kept minimal to avoid extra dependencies (like on libcudart).

jcvernaleo:
I was totally unaware of that other package.  I'll redo the PR there.  Thanks.

PR #94, opened 2017-02-07
using local magnetization instead of global

kakutsmann:
This is irrelevant as of now, since there is only one magnetization object (?), but should be safer for possible future use.


barnex:
Thanks.

PR #96, opened 2017-02-13
added missing M_sat factor in custom energies +added tests for energy

JLeliaert:
Expected values were generated by native anisotropy and exchange functions and in the tests compared to the custom one. All tests passed


godsic:
@JLeliaert Travis fails due to the inconsistencies in the Nvidia-provided CUDA repo. I will try to fix that one before merging.

PR #102, opened 2017-03-24
Jonathan

JLeliaert:
https://github.com/mumax/3/commit/d6b605ae52759a8d910fbe5b8ff0eaf43d815c30

Updated benchmarks


deltaf:
The changes for adaptive time step at finite temperature and centerbubble look very useful.
Is there a particular reason why this isn't merged?

engine/temperature.go
```diff
@@ -70,9 +70,19 @@ func (b *thermField) update() {
 		return
 	}
 
-	if FixDt == 0 {
-		util.Fatal("Finite temperature requires fixed time step. Set FixDt != 0.")
-	}
+	// after a bad step the timestep is rescaled and the noise should be rescaled accordingly
```
barnex: The time step gets re-scaled at every step.

barnex:
@JLeliaert any specific reason for closing this?

PR #103, opened 2017-04-15
Update shape.go

LLizardi:
Maybe is necessary keep the center of the cone in the center of the space


barnex:
Superseded by #104

Issue #111, opened 2017-07-04
mumax3-server aborts with error if -scan IP address contains 255


joh:
IP addresses may contain 255 but mumax3-server fails if one is passed to -scan:

```
$ mumax3-server -l 127.0.0.1:35360 -scan 10.1.255.194
2017/07/04 16:32:07 too many IP addresses to scan in10.1.255.194
```

This is caused by the byte holding the individual IP address numbers wrapping around, causing an infinite loop which fills up the IPs list. A viable fix is to change the type to uint which is able to hold numbers greater than 255, thus terminating the loop.

PR #112, opened 2017-07-04
Allow IP address to contain 255

joh:
Fix issue #111

### DIFF
--- a/cmd/mumax3-server/main.go
+++ b/cmd/mumax3-server/main.go
@@ -178,13 +178,13 @@ func parseIPs() []string {
 		if len(split) != 4 {
 			log.Fatal("invalid IP address range:", s)
 		}
-		var start, stop [4]byte
+		var start, stop [4]uint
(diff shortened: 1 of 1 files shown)


godsic:
@joh Thanks for fixing this issue. Have you actually tried to fix the atobyte function, instead of casting?

joh:
@godsic Sure, I can move the casting to the atobyte function instead if you prefer?

godsic:
@joh Sorry for not being clear. I was wondering if one can use [ParseUint](https://golang.org/pkg/strconv/#ParseUint) in atobyte(...) and then cast from uint64 to byte?

joh:
@godsic Not sure that would make sense - the atobyte function would still need to return uint.

The problem is really the for loop at cmd/mumax3-server/main.go:191 which will cause a byte to overflow after incrementing past 255. An alternative may be to change the types of A, B, C and D to uint and leave the rest of the code as it was.

PR #116, opened 2017-07-14
don't use inter exchange params at boundaries in dmi.cu

JeroenMulkers:
The inter exchange parameters (`Aex` and `Dind`) at the boundary of the geometry did depend on the exchange parameters in and out the geometry. This means that the non trivial BC for the DMI could depend on `Aex` and `Dind` outside the geometry.

This issue is solved by not using the inter exchange parameters for the BC. Instead, the exchange parameters of the cell inside the geometry are used.


godsic:
@JeroenMulkers Is it only at the material/vacuum boundary?

JeroenMulkers:
Yes. It is only a problem when the magnetization (small m) of the neighbouring cell is 0, which is the case for cells outside the geometry.

The boundaries of the grid are handled correctly due to the clamping of the index `i_`.

I've compiled with cuda 8.0. Maybe it is better to recompile it with cuda 7.5? (at the moment I do not have cuda 7.5 installed on my system)

godsic:
@JeroenMulkers No worries, there is a script in cuda folder that tries to generate PTX using broad range of CUDA SDKs to ensure greater compatibility with diverse end-user systems. If we would tag a 3.10 release, I will re-generate all the PTX files myself.

Issue #99, opened 2017-03-14
Sign issue in Sloncewski spin torque


Gh0str1d3r:
The Slonceswski spin torque in MuMax is defined as

tau = (A - alpha B) / (1+alpha^2) m x p x m + (B - alpha A)/(1+alpha^2) p x m

with

A = beta epsilon
B = beta epsilon_prime

I believe the first coefficient should be (A + alpha B). Please allow me to illustrate the issue.

The following python script determines the damping-like and field-like coefficients of the explicit LLGS, commonly called a_j and b_j, for a film thickness of 1 nm, Ms = 1e6, Lambda = 1. It just finds the field that one has to apply to keep the magnetization aligned in the out-of-plane direction.

```import subprocess
import numpy as np
import sys

def exec_command(command,verbose=False):
  '''
  Execute a command on the shell and return the standard output and
  the standard error as a result. The function will not return before
  the command has been executed and finished.
  
  Paramters
  ---------
  command : string
    Command (bash) to be executed
  verbose : bool
    It True, the command, the standard output and the standard error
    are printed to the terminal.
  
  Returns
  -------
  A tuple (standard output, standard error)
  '''
  if(verbose):  
    print(command)
  p = subprocess.Popen(
        command,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        shell=True)
  stdout, stderr = p.communicate()
  if ( stdout and verbose):
	  print('process returned stdout:\n' + stdout)
  if ( stderr and verbose):
	  print('process returned stderr:\n' + stderr)
  return stdout,stderr
  
script = """
SetGridSize(1, 1, 1)
SetCellSize(1e-6, 1e-6, 1e-9)
Msat = 1e6
Aex = 1e-11
alpha = 0.5
m = Uniform(0, 0, 1)
B_ext = vector({Bx}, {By}, 1)
DisableZhangLiTorque = True
EnableDemag = False
lambda = 1.0
Pol = {pol}
EpsilonPrime = {ep}
FixedLayer = vector(0.0, 1.0, 0)
OutputFormat = OVF1_TEXT
J = vector(0, 0, 1e12)
run(1e-9)
save(m)
"""


Bx=0
By=0
dx=0.1
dy=-0.1
signx=-1
signy=1
for i in range(20):
  with open("tmp.mx3", "w") as text_file:
    text_file.write(script.format(Bx=Bx,By=By,ep=sys.argv[1],pol=sys.argv[2]))
[...]

PR #121, opened 2017-08-18
Bugfix/slon

godsic:
mumax3 assumes the fixed layer is placed on top of the free one, which is consistent with OOMMF implementation. However there are at least two issues with this notation:

- most of the mumax3 people are not aware of this assumption (although it is clearly stated in the mumax3 paper) and use Slonczewski torque blindly;
- most of the real-world samples are made in the opposite way, i.e. fixed layer underneath the free one, similar to how Slonczewski derived his original equation.

So in mumax3 if one want to switch parallel layer a _negative_ current has to be applied, while in Slonczewski paper a _positive_ current is needed. 

To avoid this confusion, I propose to introduce a FixedLayerPosition quantity that might have one of the following values:
FIXEDLAYER_TOP (default) which does not change how mumax3 behaves, i.e. one needs negative current to switch parallel layers;
FIXEDLAYER_BOTTOM which flips the signs of both damping-like and field-like terms, so that positive current is needed to switch parallel layers.


engine/torque.go
```diff
@@ -33,6 +34,9 @@ func init() {
 	DeclVar("DisableZhangLiTorque", &DisableZhangLiTorque, "Disables Zhang-Li torque (default=false)")
 	DeclVar("DisableSlonczewskiTorque", &DisableSlonczewskiTorque, "Disables Slonczewski torque (default=false)")
 	DeclVar("DoPrecess", &Precess, "Enables LL precession (default=true)")
+	DeclLValue("FixedLayerPosition", &flposition{}, "Position of the fixed layer: FIXEDLAYER_TOP, FIXEDLAYER_BOTTOM (default=FIXEDLAYER_TOP)")
+	DeclROnly("FIXEDLAYER_TOP", FIXEDLAYER_TOP, "FixedLayerPosition = FIXEDLAYER_TOP instructs mumax3 that fixed layer is on top of the free layer")
+	DeclROnly("FIXEDLAYER_BOTTOM", FIXEDLAYER_BOTTOM, "OutputFormat = OVF2_BINARY instructs mumax3 that fixed layer is underneath the free layer")
```
barnex: copy-pasta typo: OutputFormat = OVF2_BINARY

engine/torque.go
```diff
@@ -98,6 +105,7 @@ func AddSTTorque(dst *data.Slice) {
 		defer epsPrime.Recycle()
 		cuda.AddSlonczewskiTorque2(dst, M.Buffer(),
 			msat, j, fixedP, alpha, pol, lambda, epsPrime, Mesh())
+		j.SetMul(Z, jmul_z)
```
barnex: redundant because of previous defer j.SetMul... ?

godsic:
@barnex thanks for the feedback. It turned out that if current is set from file, the multiplier got overridden when cuda kernel in invoked. So I have re-worked this pull request to solve this issues. Please check it out. There will be a separate pull request to fix the sign of the pre-prefactor in the Slonczewski damping-like torque.

godsic:
before 53dbec4 there was a significant discrepancy between OOMMF and mumax3 when un-physically large value of the non-adiabaticity is used (as set by the secondary spin-torque parameter):

before-fix

As it has been pointed out in #99 the problem is in the pre-factor of the damping-like term. So 53dbec4 flips the sign as suggested in #99 so that we get a perfect agreement with OOMMF:

fix

I would not expect any major harm from this issue, since with realistic values of the secondary spin-torque parameter the effect is at least of the second order.

PR #129, opened 2017-09-22
Add ext_make3dgrains and testfile

peytondmurray:
Added an extension which allows for 3d Voronoi tesselation. It allows the user to choose grain size, specify which region numbers to use for the tesselation, as well as the specific shape inside your simulation space to tesselate. I included an example where part of the space is broken into Voronoi cells, and another part is left as a single uniform layer.

### DIFF
--- a/engine/ext_make3dgrains.go
+++ b/engine/ext_make3dgrains.go
@@ -22,7 +22,6 @@ func Voronoi3d(grainsize float64, startRegion int, numRegions int, inputShape Sh
 type tesselation3d struct {
 	grainsize   float64
 	maxRegion   int
-	seed        int64
 	rnd         *rand.Rand
 	startRegion int
 	shape       Shape			//Shape of the tesselated region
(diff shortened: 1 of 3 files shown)


engine/ext_make3dgrains.go
```diff
@@ -0,0 +1,158 @@
...
+// Stores location of each cell
+type cellLocs struct {x, y, z float64}
+
+// nRegion exclusive
+func newTesselation3d(grainsize float64, nRegion int, seed int64, startRegion int, inputShape Shape) *tesselation3d {
+	t := tesselation3d{grainsize,
+		nRegion,
+		//make(map[int3d][]center3d),
+		seed,
+		rand.New(rand.NewSource(0)),
```
barnex: shouldn't seed be passed to NewSource?

peytondmurray: Yes, good catch. I have removed the seed from the tesselation3d type, and just passed the seed directly to NewSource.

barnex:
Cool, thanks!

1eevy:
Good extension! Is it possible to create grain boundary layers between the 3dgrains by ext_make3dgrains? Can we create 3dgrains with core-shell structures? Thanks!

PR #138, opened 2017-10-19
Log output from failed jobs

### LINKED ISSUE #137: No output from failed jobs when running multiple in queue
When running multiple jobs in queue, and a job fails, its output is discarded. This makes debugging failed jobs difficult.

joh:
When running multiple jobs in queue, log the output of failing jobs.

Fixes #137


barnex:
Thanks.

PR #139, opened 2017-11-09
Aleste2 custom extensions

aleste2:
Change List
-------------
Increased the possiblity of max regions up to 8192 but mantained to 256 due to important speed decrease.
Inluded ShiftY similar to Shift for XY shifting.
Calculates Skyrmion position and radius.
Non adjacent interlayer exchange (non local RKKY)
LLB solver.
LLB+Joule Heating + External heating solver.
Independent Slonczewki Torque parameters if needed (SOT and STT).
Preliminary current solver (uniform resistivity).

Probably a lot of bugs and horrible coding.
Not sure about how github works...


godsic:
@aleste2 wow, that's a lot of stuff to digest. Would you be so kind to split it up into smaller pull requests, i.e. one per specific feature?

aleste2:
Ok, tell me what is the best way to do it. To put you into context:
I have derived my code from 3.9.3c, so some of your last modifications are not included. Some comments about each part:
- Increasing the maximum number of regions over 256: it modifies almost all the code, as the variable type changes from 8 bit to 16 bit, and it affects multiple files (go and cuda). I should check it if has some important effect in speed (only the type, maintaining the limit to 256; increasing it to 8192 clearly decreases the speed). If really needed I can implement it, but it is a lot of work if I have to do it in your last version. Anyway I have it in my code and I am running with the 256 limit and compiling specific versions when a high number of regions is needed. Let me know what you think about it.
- ShiftY: I think it is already implemented in the actual master branch. Am I correct?
- RKKY for non adjacent layers: the same already included in the actual master branch. Maybe different implementation.
- LLB solvers: that is the more interesting contribution from my part. I have only to modify some parts of the code if byte/word conflicts appear. Also I have to change the solver number to skip conflicts with the last one included in the master branch.
- Skyrmion position and radius: only one file, it is just postprocesing, easy to include.
- Independent Slonczewki Torque parameters if needed (SOT and STT): also quite easy to introduce, but maybe not much interesting (only for complicated current shapes).
- Current solver: although not much accurate due to finite differences method limitations (external finite element solver preferred) can be useful to generate a mask for the current. It is again separated from the rest of the code, only new files included.
(thats a lot of information)
So tell me, what is the best way to organize the work:
- Download your actual master branch, and implement, for example, the LLB solvers as a first contribution? and then make the same with separate branches for the other parts?
or
- Download your actual master branch, include the increase of the number of regions and then include the other parts. Not sure if this is interesting for many people.
Sorry for the extension...
Thanks.

godsic:
@aleste2 There are indeed some conflicts with the masters's tip. To simplify the review process I would suggest you to (a) split any specific feature in a separate pull request and (b) rebase your fork(s) against master. @JLeliaert @JeroenMulkers Correct me if I am wrong, but  "ShiftY" and "Skyrmion position and radius" are indeed in the master. If so, then @aleste2 should probably omit them in his pull requests.

aleste2:
Ok you can close this pull request and I'll start a new one with LLB solvers based in master branch. It will take me some days to complile and check everything.

godsic:
@aleste2 Good, looking forward to hearing from you.

PR #140, opened 2017-11-13
LLB solver + Thermal equation (joule and external heating)

aleste2:
LLB equation and thermal equation.
To do: PBC (not used)


aleste2:
Added two missed files.

https://github.com/aleste2/3/commit/03cb4de111a6a129ba9af131551a0acb806ac19d

barnex:
Seems like this needs some more work:
  * files need correct formatting. run gofmt.
  * please use lowercase file names
  * code does not compile, kernels missing?
  * doc comments missing or not properly formatted
  * no test files
  * this should probably be an extension for the time being. I.e. start all files, quantities, functions with ext_ to indicate to users this is not battle-hardened just yet.

kkingstoun:
@aleste2  will you publish your branch with an extended number of regions?

aleste2:
Not by now, it is a lot of work to update it to the master branch, as it was based on 3.9.3, and it impilies an important speed decrease.

kkingstoun:
But could you send me your version based on 3.9.3?

Issue #146, opened 2018-01-10
Relax() stopping criterion


jsampaio:
The relax() stopping criterion seems wrong to me. The [current relax() procedure](https://github.com/mumax/3/blob/3.10/engine/relax.go) is:
 1. Run until the total energy stops decreasing.
 2. Run until the average torque stops decreasing. 
 3. Repeat step 2, decreasing MaxErr until MaxErr<=1e-9.

There are some issues that seem wrong to me:
 - The torque criterion should consider the maximum torque and not the average torque. Often, only a very small subset of the spins are moving, and relax() should keep running until those spins relax. The maximum torque will measure those fastest spins, independently of the number of already relaxed spins. Moreover, the average torque is noisier, as it's a sum of lots of near-zero values.
 - The torque criterion should not be that the torque stops decreasing, it should be that the torque nearly reaches zero (within the numerical noise). A slowly relaxing system may show a constant (non-zero) torque for a long time, such as a domain wall moving in a small constant field. Also, in many relaxation situations the torque may increase, before the system is relaxed. For example, a depinning domain wall at its critical field will start by deforming, slower and slower, up to a point when it depins and the torques increase -- relax() should not stop in this case.
 - if the initial MaxErr is <=1e-9, the step 2 will not run.

I suggest that the maximum torque be taken instead of the average torque, and that relax() stop when maxTorque reaches a user-definable threshold. I'm not very familiar with the code, but something like changing lines 62-76 of [relax.go](https://github.com/mumax/3/blob/3.10/engine/relax.go) by:
```
    var thresholdTorque := 1e-3; //user-settable
    maxTorque := func() float32 {
        return cuda.MaxVecNorm(solver.k1)
    }
     for maxTorque() < thresholdTorque && !pause {
            relaxSteps(N)
        }
    for MaxErr > 1e-9 && !pause {
        MaxErr /= math.Sqrt2
        for maxTorque() > thresholdTorque && !pause {
            relaxSteps(N)
        }
    }
```
ed: I've written the code implementing these changes in pull request #148 .

barnex:
Although the points by @jsampaio are valid in principle, they are nearly impossible to do well in practice. The reason is that E_total and maxtorque are simply too noisy. The current implementation, although not perfect, is a least robust against noise and yields the correct results when compared to OOMMF, _for a wide range of material parameters and time scales_.

[...]

PR #148, opened 2018-01-11
Relax() stopping criterion (issue #146).

jsampaio:
Code for  issue #146 . This changes the relax() stopping condition so that it stops when the maximum torque reaches a user definable threshold ('RelaxTorqueThreshold').


jsampaio:
@godsic This pull request changes the Relax() current behaviour. As it is, relax() does not stop when the system is relaxed, it stops when the system is steadily evolving, i.e. when the dm/dt is constant. The main change of this patch is changing the stopping condition to a dm/dt approaching zero. I've written more details of the things in the current Relax() that I think should change in issue #146 .

godsic:
@jsampaio thanks for the clarification. Our merging policy is that any changes, apart from the bug fixes, shouldn't change mumax3 default behavior. So I would suggest you to use either RelaxTorqueThreshold = 0. or RelaxTorqueThreshold = -1. as the default value (whatever value feels more logical to you) that reproduces Relax() current behavior. Then, if desired user can override this by setting RelaxTorqueThreshold to the desired value.

godsic:
@jsampaio Thanks. I guess we are approaching the point of merging. I basically have three comments:
- at the moment RelaxTorqueThreshold = -1. and RelaxTorqueThreshold = 0 are equivalent, both will trigger the default behaviour. Is it intentional? 
- to minimize possible regressions please try to keep the mumax3 original code as it is.
- have you tested your changes against unit tests?

JeroenMulkers:
The Relaxation procedure is an essential part of mumax3, so I think it's best not to be too hasty in modifying anything.

Consider the following statement:

Stop the relaxation if the 
    (1) average
    (2) maximum
of the torque 
    (a) doesn't change any more
    (b) is approximately zero

Basically, we have a '2D binary option space'. In the current version we use option (1a), which should indeed be the default if we add other options in order to ensure BC.

If we add option (2b) as suggested by @jsampaio, I think we should include the other options as well. That being said, I am not yet convinced that adding option (b) is necessary. @jsampaio, do you have a script which clearly demonstrates that (2b) is a better stopping criteria than (1a)?

jsampaio:
@godsic : I still think that the default (current) behaviour of relax() is not desirable, as it frequently yields results very far away from a relaxed state. Perhaps if you guys agree, in the future, we can change the default RelaxTorqueThreshold so to use the new relax() condition, if we're all satisfied that it poses no new problems.

I have a demonstrator simulation, comparing the old and the new relax, and Minimize() : [relaxTriangularTrackSimplified3.txt](https://github.com/mumax/3/files/1639695/relaxTriangularTrackSimplified3.txt)


In the current code, any value of RelaxTorqueThreshold <=0 will use the default behaviour. I think this is ok, as the threshold for the maximum torque should always be a positive finite number.

I've run the tests on the folder /test.  However, I get some errors with both branches of the code, on tests that do not have relax()...

JLeliaert:
I agree with @jsampaio that the current relax function sometimes does not lead to an energy minimum and an improved version is welcome. Even if we want to keep the default behaviour as it is now, I think we should include the new version with the RelaxTorqueThreshold option.

jsampaio:
@JeroenMulkers you've summarised the problem very neatly. My opinion is that option (2b) is better than the others, if we want that Relax() drive the system to a local energy minimum. Let me explain what I think is bad with the other options:

* The average torque (options 1a 1b) suffers from numerical noise. Near equilibrium, most spins will be static, and so the average will be a sum of a large number of near-zero values. This means that Relax() would stop even if there was still one spin moving quickly, as long as there's also lots of static spins. Moreover, the final state would change with the total number of spins in the system, as a bigger system would have more static spins adding noise to the average.

* Waiting for the torque to be constant (options 1a and 2a) is not a condition for a local energy minimum. A system that is steadily evolving, but still far away from the energy minimum, can show a constant torque. 

There are lots of cases where the current Relax() (with condition 1a) stops too early. I've attached an example in the previous comment: a domain wall in a wedged track.

JeroenMulkers:
@jsampaio thank you for the script. It will be definitely useful to improve the relax function!

I agree with you that a converging criteria based on the maximum of the torque instead of the average torque might be better in some cases.

I also want to point out that **if you have a constant torque, then this constant has to be zero** because the magnetization can only make rotations on the unit sphere. This means that it is a good condition to check if the magnetization is in a local energy minimum.

jsampaio:
@JeroenMulkers : That's only true for the torque vector. In the case of relax(), if there's a constant torque, that constant can be different from zero. It's not the torque vector we test, it's the magnitude of the torque vector. Moreover, we consider either the maximum or the average of the torque over all the cells. 

Take the case of a moving domain wall at constant speed. The maximum torque (magnitude) will be constant but correspond to different cells as the DW moves. The average torque (magnitude) will also be constant.

JeroenMulkers:
I see. It is indeed the magnitude of the torque which is checked and which could be a constant. Thank you for pointing that out!

godsic:
@jsampaio Would you be so kind to replace lines 85-95 in your most recent relax.go implementation with **exact** copy of the master's code and I am happy to merge then.

PR #152, opened 2018-01-25
allow to combine arrows with scalar colormaps 

godsic:
Before this change mumax3-convert would normally output either:

m_xrange192-320_yrange192-320_012500

or 

m_xrange192-320_yrange192-320_012500

After this change one can do something like this:
m_xrange192-320_yrange192-320_012500

### DIFF
--- a/cmd/mumax3-convert/main.go
+++ b/cmd/mumax3-convert/main.go
@@ -82,7 +82,7 @@ var (
 )
 
 var (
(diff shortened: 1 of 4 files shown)


godsic:
@barnex not sure it is done in idiomatic way, but seems like this is the most compact version.

barnex:
Thanks.

PR #156, opened 2018-02-06
ext_CenterWall per region

jsampaio:
This change adds ext_centerWallInRegion(R,c), which works like ext_centerWall but only for a specified region. This is useful for simulating SAFs, for example.


godsic:
@jsampaio would you be so kind to clean this up?

godsic:
@jsampaio Also, have you tried to run unit tests?

godsic:
@jsampaio if you want this merged, then please cleanup and resolve conflicts.

jsampaio:
@godsic : I'll look into it again. I think the changes I did to add ext_centerWallInRegion(R,c) got mixed in with other (unrelated) changes to the code...

jsampaio:
@godsic I cleaned it up a bit and added a new option for shifting the magnetisation that uses the current values at the border for the new cells introduced by shift. This mode is more useful for simulations of antiparallel coupled layers (SAF like in the RKKY example). Because of the clean-up, I had to do a new branch, and so I'll do a new pull request.

PR #159, opened 2018-02-09
Added benchmark numbers for 3 Tesla cards

Dmytro-Apalkov:


kkingstoun:
@godsic could you update graph from the main webpage with this new results?

PR #161, opened 2018-02-25
cache MFM kernels

JLeliaert:
cache MFM kernels, similarly to demag kernels


barnex:
Re @godsic: the kernel is several times larger than a magnetization file. However, one does not necessarily need to use /tmp.

JLeliaert:
Each MFM kernel file has the same size as the demag kernel file of the same
system. For a 1000 000 cell problem, this is roughly 4 Mb.

PR #194, opened 2018-09-21
Collapsible job lists

jsampaio:
This changes the mumax-server status page so that the job lists are collapsible, in the style of the elements of the mumax3 simulation page. This way, you can find your simulations more quickly, without having to scroll through the long job lists of other users.


barnex:
Very cool.

PR #195, opened 2018-09-27
add openbc

JeroenMulkers:
A few notes about the implementation of open boundary conditions:

- The default is "openbc = false". So the default behaviour matches the old behaviour.
- The Neumann BC and open BC are equivalent for non-chiral systems. Hence, I only implemented open BC in the dmi and dmibulk kernels.
- The difference of relaxed magnetization configurations with Neumann and open BC converges to zero for a decreasing cell size.
- test/dmienergy-vardmi.mx3 and test/dmienergy.mx3 fail when using open BC. By making the error tolerance a bit less strict, also these tests succeed.

Conclusion: so far it seems to me that the difference between open and Neumann BC is rather small and, for relaxed magnetizations, becomes even smaller when one decreases the cell size. 

(kernels are not yet recompiled)


godsic:
@JeroenMulkers Travis fails, could you please rebase against master and double check it compiles?

JeroenMulkers:
Rebuild the kernels. Should be OK now.

JLeliaert:
Travis succeeds

godsic:
@JeroenMulkers Are there any tests of openbc in particular?

JeroenMulkers:
Because the difference between the BC is so small, I don't really know how
to create a non-trivial test for the open BC.

It is certainly possible to make the choice of the BC dependent on the
axis. However, I do not see the point. If users are convinced that open BC
should be used, then they should use them at every boundary.

godsic:
@JeroenMulkers can we benefit from these tests https://github.com/fangohr/paper-supplement-standard-problem-dmi/tree/master/sims ?

godsic:
Can we benefit from these tests https://github.com/fangohr/paper-supplement-standard-problem-dmi/tree/master/sims ?

JeroenMulkers:
I added a test for the open BC based on the '1D' standard problem proposed by Cortés-Ortuño et al. See test/openbc.mx3 for more details.

godsic:
@JeroenMulkers I am more comfortable with 1e-5 tolerance. Cam you achieve it with smaller cells? Then, would you be so kind to make three identical tests, but aligned along x, y and z?

JeroenMulkers:
I could bring the tolerance down to 1e-3. The analytical result gives the canting at the edge of an infinitely long wire, whereas we of course simulate a finite wire. Therefore, it is hard to let the simulated canting converge to the analytical one.

It is easy to shown that a 1D model can be simulated by using one row of cells (let's say in the x direction) and periodic boundary conditions in the y direction. Hence, we can also compare the case with open and periodic boundary conditions. In this case, a tolerance of 1e-5 can be achieved.

I have added the case where the nano wire is aligned along the y direction. There is no canting at the edges of a nano wire along the z direction, so this case is not tested.

PR #196, opened 2018-10-01
fixed cuda version statement

JLeliaert:
When compiling the binaries for older CUDA versions I noticed that mumax does not print the CUDA version actually used, but that cu_GetDriverVersion, actually prints the highest CUDA version supported by the driver. Was this intentional? If not, these changes allow mumax to print the actual CUDA version.

### DIFF
--- a/cuda/cu/version.go
+++ b/cuda/cu/version.go
@@ -3,14 +3,17 @@ package cu
 // This file implements CUDA driver version management
 
 //#include <cuda.h>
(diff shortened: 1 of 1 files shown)


godsic:
Why are there xuebing commits as well?

godsic:
@JLeliaert lets split the pull requests, so could you please reset to 2ae01e4a7371e06f7da3c3db0cdeab074a8f052b ?

PR #202, opened 2018-10-10
Add Helical initial magnetisation function

rpep:
Hi there,

This function just adds a function which allows a helical initial magnetisation propogating along (1, 0, 0). I had some trouble with gofmt not working on the machine I compile on, so I had to disable it - if there are any changes that I'd need to make to get this accepted then please let me know.

Best wishes,

Ryan Pepper
University of Southampton

### DIFF
--- a/engine/config.go
+++ b/engine/config.go
@@ -18,7 +18,7 @@ func init() {
 	DeclFunc("VortexWall", VortexWall, "Vortex wall magnetization with given mx in left and right domain and core circulation and polarization")
 	DeclFunc("RandomMag", RandomMag, "Random magnetization")
 	DeclFunc("RandomMagSeed", RandomMagSeed, "Random magnetization with given seed")
-	DeclFunc("Helical", HelicalMag, "Helical magnetization with with q-vector along (1, 0, 0) and helical length Ld")
+	DeclFunc("Helical", HelicalMag, "Helical magnetization with helical length Ld and with q-vector along (qx, qy)")
 }
 
 // Magnetic configuration returns m vector for position (x,y,z)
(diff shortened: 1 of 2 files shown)


JeroenMulkers:
This is definitely useful, but it might be even better if we could generalize this so that it works for ever (3D) wave vector.

rpep:
OK, let me edit it and come back to you on that, shouldn't be too difficult. I'll put together an example as well.

rpep:
I've just tried messing around locally on my own machine; I can't currently build against the Master branch even after upgrading to CUDA 10.0 - I get the error:

> No binary for GPU. Your nvidia driver may be out-of-date

Any ideas? I'm using Ubuntu 18.04

kkingstoun:
You have to update the driver.

rpep:
Not sure what I previously as I've now uninstalled it in trying to fix it; are you guys building against drivers in a PPA for Mumax generally, or from the binary distributions on NVidia's website?

rpep:
The card is a 750 Ti, so fairly old, although it's worked fine up til now.

godsic:
The card is still supported. I would strongly suggest you to setup NVidia own repository as it provides both cuda and appropriate driver.

rpep:
I've just done that; seems to work fine now - thanks a lot. I'd used the local runfile previously and I suppose there were some conflicting files lying around.

godsic:
@rpep Please let us know when the pull request is ready to go.

JeroenMulkers:
@rpep Small note: when you specify the wave vector q, you do not have to specify the wavelength L since q = 2*pi/L. 

Your idea to add the helical state to the available magnetization configurations, inspired me to implement also the general conical configuration (this includes the helical state, cycloidal state, and anything in between). See commit 93e50f6.

I believe it is good practice to implement only the general case (conical state) and make a user friendly wrapper for the special case (helical state). Please, have a look at commit 93e50f6 and let me know if you agree?

rpep:
Aha, I was thinking about that too. I wasn't sure whether it was a better interface for people to think about the helical length separate to the direction, because it's somewhat natural for people to state things like [110], [101], etc in papers and that would line up more consistently with that.

I agree; and your way looks much simpler than my way! I'll close this now; thanks!

PR #206, opened 2018-10-25
TEMP path

kkingstoun:
Hi,

this function allows to detect os, and set the correct path to the temp directory. This code will help mainly windows users.

Kking


godsic:
Could you please rebase against master?

JeroenMulkers:
A few nitpicks:

- Why are there changes in .gitignore and  bench/gpus.txt in this pull request?
- The name of the function "DetectOs" is not ideal. The function returns the tmp directory and not the OS.

I think we can solve the issue of the tmp dir (also mentioned several times on the mailing list) simply by changing the default value of the cache flag from \tmp to ['Os.TempDir()'](https://golang.org/pkg/os/#TempDir). @godsic Do you thing this is a safe solution?

kkingstoun:
@JeroenMulkers I wanted to publish only this file but I had a few problems with Github ... I have to learn more about git :) 

It works fine for Windows and Ubuntu, but I did not have time to check on different distributions on linux.

JeroenMulkers:
@kkingstoun , thank you for your suggestion to use the os.TempDir(). The issues with the cache directory are resolved in 3e5122d.

PR #216, opened 2019-01-26
Floating magnifier showing details in the result image

jsampaio:
This adds a magnifier to the html interface of the mumax results page. The magnifier floats on top of the image and shows a zoomed-in detail of the image under the cursor.

### DIFF
--- a/engine/html.go
+++ b/engine/html.go
@@ -28,6 +28,77 @@ const templText = `
 	    }
 	</script>
 
+
+
+	<style type="text/css">
+		* {box-sizing: border-box;}
+		.img-loupe-div { position: relative; }
+		.img-loupe {
+		  position: absolute;
+		  border: 1px solid #000;
+		  border-radius: 2% 50% 50%;
+		  box-shadow: 3px 3px 8px rgba(0, 0, 0, 0.4) ;
+		  cursor: crosshair;
+		  width: 150px;
+		  height: 150px;
+		  display: none;
+		}
+	</style>
+	<script type="text/javascript">
+		// image Loupe
+		function updateLoupe(imgID, zoom) {
+		var img, loupe;
+		img = document.getElementById(imgID);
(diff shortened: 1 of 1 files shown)


godsic:
@jsampaio awesome, thanks! have you tried to test it against different browsers?

jsampaio:
@godsic  :  thanks! I've tried on safari and firefox.

godsic:
@jsampaio how the image of the loupe is constructed? Is it fetched from elsewhere?

jsampaio:
It’s the same image, from the same address. The zoom is done just with css and JavaScript in the user’s browser

kkingstoun:
Hi,

I made it in a little bit different way:

https://www.youtube.com/watch?v=xN2JIwht92g&feature=youtu.be

Then I plan to add zooming functionality using wheel in mouse. What do you think about this solution?

jsampaio:
My solution or the two from @kkingstoun 's seem ok to me. My comments on the three solutions:

- The loupe (my solution) has the advantage that the user immediately understands that there is a way to zoom in, and it works immediately; there's no learning to be done, no clicking to be remembered.

- The scaling by dragging (the solutions from @kkingstoun ) has the advantage that once the user has dragged the image, they can see it entirely. However, it is not immediately clear that it is possible to do it, and it demands a bit more interaction. If you choose this, I think there should be some indication (textual or graphical) that it is possible to scale the image by dragging.

- On the second video, the dragging changed the image aspect ratio. I think the aspect ratio should be kept constant.

kkingstoun:
@jsampaio you have the right suggestions, I will prepare graphical toolbox with icons representing loup, drag, resize, and zoom. I think that user, should have an option to use it, because not every time the loup is useful. I'm thinking also about requesting render to return an image in higher resolution (upon request) for loop and zooming, but i have to test how it will influcene on the computing performance.

PR #221, opened 2019-03-10
Feature/shift : copy the border cells when shifting M

jsampaio:
this change introduces a new option for the shift. Instead of filling the new cells introduced by shift with the values of ShiftMagL and ShiftMagR, it is now possible to use the current values at the border by setting `FudgeShift = true`. 

It is still needed to set `ShiftMagL`/`R` because they're used to determine the shift direction (as is the previous behaviour).


godsic:
@jsampaio Could you please clean-up this pull request to only contain `FudgeShift` and provide unit tests?

jsampaio:
@godsic : it's been a while since I proposed a change to the code, so I have to see how that is done... 

A unit test is a script that tests if the function is still working, like e.g. "test/centerwall.mx3" ? 

To propose a cleaned-up version, should I make a branch of the current mumax base on the mumax git? Every time I did a new feature on a branch of my own fork, it gets mixed up with all the other changes (like now...)

godsic:
@jsampaio 

Yes, you can mimic that test.

Yes, please make a fork of master, then apply only `Feature/shift` stuff and make a pull request again.

jsampaio:
This pull request is replaced by #316 (the same code, but in a clean branch).

PR #258, opened 2020-05-15
Update topologicalcharge.go

yuli-nano:


JeroenMulkers:
This seems to be the same as pull request 257

PR #265, opened 2020-07-02
Feature/topcharge

JeroenMulkers:
This is an alternative for the `topologicalchargelattice` implementation proposed in pull request #262.

The local topological charge is computed by taking a weighted sum of the topological charge contributions of the four elementary triangles defined by the four nearest neighbours. 

In general, the elementary triangles of neighbouring cells overlap, in which case the charge contributions are counted twice. This is solved by using a weight of 1/2. (Exactly the same is done in the topologicalchargelattice implementation in #262)

However, the triangle at a convex corner of the geometry does not overlap any other triangles. Hence, at convex corners we use a weight of 1.


godsic:
@JeroenMulkers How does it score accuracy-wise in tests?

JeroenMulkers:
It gives exactly the same results as Joo-Von's implementation in the unit tests. These unit tests have a trivial geometry, in which case both implementations are equivalent, and hence we expect that it yields the same results

The only thing which is different here, is that the contribution of a triangle is counted twice in a convex corner of the geometry.

godsic:
@JeroenMulkers Good. Please give credit to the Joo-Von in the sources and make sure we ask to cite his paper if feature is used.

PR #271, opened 2020-10-14
Feature/Time series signal from file support

JakeLove:
Added the capability to load time series data from external files so that material/region properties can be arbitrarily set to different values throughout the simulation. Currently to vary a property such as the external field you could write something like:

`
B_ext = A*sin(t)
`

Which is nice if you want B_ext to be an analytical expression, however this wasn't sufficient for my case where I need material properties to vary based on datasets from another source.

The functionality I have added allows you to load a multicolumned file into mumax3, then use each column as a `ScalarFunction` of (typically) `t`.

Example input file is given below showing all the exposed functions and an example usage.

```
// Units
ns := 1e-9

// A multidimensional signal can be loaded from a file containing an arbitary number of numerical columns. A timestep
// is then given that defines the sample rate of the signal (the time between adjacent rows).
S := LoadSignal("file.txt", 2*ns)

// Example 'file.txt':

//  0.0  1.2  2.8
//  1.1  2.7  3.8
//  1.9  1.3  3.5
//  1.5  1.3  4.9
//  4.7  3.4  2.3
//  5.3  2.3  1.0
//  1.0  2.2  0.8
//  1.3  3.7  0.8
//  4.1  4.3  1.5

// Various properties of the signal can be gathered with 
dt         := GetSignalDt(S)
len        := GetSignalLength(S)
duration   := GetSignalDuration(S)
dimensions := GetSignalDimensions(S)

// or dt can be set to a new value to repeat an experiment with a different sample rate
SetSignalDt(S, 1*ns)

// Signal values can be accessed by using SignalValue as a 'ScalarFunction' type, where the last argument represents
// the column index to read from
tableaddvar(SignalValue(S, t, 0), "external_field", "mT")
tableaddvar(SignalValue(S, t, 1), "vcc", "mV")
tableaddvar(SignalValue(S, t, 2), "vss", "mV")

// Intended usage is something like below:
//   B_ext = SignalValue(S, t, 0))
//   AppliedPotential.setRegion(vcc_region_id, SignalValue(S, t, 1))
//   AppliedPotential.setRegion(vss_region_id, SignalValue(S, t, 2))


tableautosave(0.01*ns)

// Arbitary system to get mumax3 running...
SetGridsize(256, 256, 1)
SetCellsize(3e-9, 3e-9, 3e-9)
Msat  = 800e3
Aex   = 13e-12

run(10*ns)
```

First time using GoLang so let me know if anything should be more 'Go-esque'.

### DIFF
--- a/engine/signal.go
+++ b/engine/signal.go
@@ -1,0 +1,120 @@
+package engine
+
+import (
+    "os"
+    "math"
+    "strings"
+    "bufio"
+    "strconv"
+)
+
+func init() {
+    DeclFunc("LoadSignal", LoadSignal, "...")
+    DeclFunc("SignalValue", SignalValue, "...")
+    DeclFunc("SetSignalDt", GetSignalDt, "...")
+    DeclFunc("GetSignalDt", GetSignalDt, "...")
+    DeclFunc("GetSignalLength", GetSignalLength, "...")
+    DeclFunc("GetSignalDimensions", GetSignalDimensions, "...")
+    DeclFunc("GetSignalDuration", GetSignalDuration, "...")
+}
+
+type Signal struct {
+    dt      float64
+    len     int
+    dim     int
+    data    []float64
+}
+
+func GetSignalDt(S* Signal) float64 {
+    return S.dt
+}
+
+func GetSignalLength(S* Signal) int {
+    return S.len
+}
+
+func GetSignalDuration(S* Signal) float64 {
+    return float64(S.len) * S.dt
+}
+
+func GetSignalDimensions(S* Signal) int {
+    return S.dim
+}
+
+func SetSignalDt(S* Signal, dt float64) {
+    S.dt = dt
+}
+
+// Credit to: https://stackoverflow.com/questions/5884154/read-text-file-into-string-array-and-write
+func ReadLines(fname string) ([]string, error) {
+    file, err := os.Open(fname)
+
+    if err != nil {
+        return nil, err
+    }
+
+    defer file.Close()
+
+    var lines []string
+
+    scanner := bufio.NewScanner(file)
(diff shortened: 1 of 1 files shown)


JeroenMulkers:
Hi @JakeLove . Having the option to define a (time) signal based on a data file could be very useful for many mumax3 users. 

I think the code is nicely written but I do have some small suggestions on the coding style:

- The 'GetFunctions' could be replaced by [go methods](https://tour.golang.org/methods/1) to make it more 'Go-esque'. If you do this, then you only have to declare the "LoadSignal" function in init(). 
- Unexpected behaviour could occur if the number of fields on a line in a datafile does not match the dimensionality of the signal (i.e. the number of fields on the first line).
- There is no need to flatten the data field in Signal. It makes more sense to use a 2D slice (slice of slices/arrays) here. 
- I would remove the len and dim fields from Signal. These values can be deduced from the data slice dimensions.

Some other remarks:

- The examples with the appliedpotential are out of place here because the Poisson solver is not included in the mumax3 master branch.
- I don't think that "B_ext = SignalValue(S, t, 0))" will work because SignalValue returns a float and not a vector.
- Instead of defining the time dependency (SetSignalDt) in the mumax3 script, I think it would make more sense to read the time values from the data file as well. In this case, the time signal is fully defined by the datafile and not split over the datafile and the inputscript. Furthermore, it makes it possible to also use unevenly spaced time series as a Signal.


I implemented a similar feature some time ago myself (actually specifically for our joined project). You can find the code [here](https://github.com/mumax/3/blob/feature/functionfromfile/engine/functionfromfile.go). We can use either this implementation or we can use your implementation (after taking the above comments into account). Let me know which implementation you prefer and why.

PR #272, opened 2020-10-19
Feature/gonumplot

JeroenMulkers:
This pull request changes how the plot in the web gui is created. Instead of calling gnuplot, the gonum/plot package will be used to create the plot.

The **advantages** of using gonum/plot over gnuplot are:
* No need to spawn a subprocess
* It removes the gnuplot dependency

### DIFF
--- a/engine/html.go
+++ b/engine/html.go
@@ -265,14 +265,16 @@ m = {{.Data.Configs | .SelectArray "mselect" "Uniform"}} {{.TextBox "margs" "(1,
 
 
 
-{{.Data.Div "gnuplot"}}
+{{.Data.Div "plot"}}
+
 <p title="{{$.Data.Doc "TableAutoSave"}}">
 	TableAutosave: {{.TextBox "tableAutoSave" "0" }}  s 
 </p>
-		Plot of "table.txt", provided table is being autosaved and gnuplot installed.<br>
-		<b>plot "table.txt" using {{.TextBox "usingx" "1"}} : {{.TextBox "usingy" "2"}} with lines </b><br/>
-		<p class=ErrorBox>{{.Span "plotErr" ""}}</p>
-		{{.Img "plot" "/plot/"}}
+<p>
+	Plot columns {{.TextBox "usingx" "0" "size=1"}} and {{.TextBox "usingy" "1" "size=1"}} of "table.txt".
+</p>
+<p class=ErrorBox>{{.Span "plotErr" ""}}</p>
+{{.Img "plot" "/plot/"}}
 
 </div>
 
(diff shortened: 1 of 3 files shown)


godsic:
Looks neat, but I will give it a closer look.

godsic:
@JeroenMulkers 
I've noticed a few things that alarmed me a bit
- `gonum's plot` pulls in many dependencies like `golatex`, `gopdf`, etc. My concern is that those might call some binaries that might not be available on the client side, in particular, on Windows machines. Could you please test on Windows and double check it works as expected?
- I've noticed that for long running simulations I frequently get `broken pipe` errors if I open web interface on a remote machine. Maybe the figure update interval is a bit too short?
- The default font is a bit too academic and contrasts with most browsers defaults. Gnuplot hardcoded `Arial`, but `Sans`-like fonts should also work.
-  I would prefer to have axes on all sides as it sometimes simplifies reading the figure data.

kkingstoun:
I can confirm that the gonum's plot works very well on Windows. I have also seen several times "broken pipe" errors.

godsic:
@JeroenMulkers In addition, I would suggest to not add brackets to labels if units are empty.

JeroenMulkers:
1. The update interval of the plot can not be set explicitly. A possible solution is to cache to plot figure and give it a certain expiration date (e.g. 1s).  This is what is done in commit #24c775e.
 
2. I also changed the data format of the image from svg to png because the svg file can become quiet large (~Megabytes) when the amount of data increases, whereas a png file remains small (kilobytes).

These two changes drastically reduce the number of broken pipe errors. There are still disadvantages of using gonum/plot over gnuplot:
- The line plots do not seems to be always correct if the plot is written to a png file. For this reason, only a scatter plot is shown
- Lines are plotted correctly in the svg files, but these can become quiet big and do not seem to support sans serif fonts.
- As mentioned by @godsic , the gonum/plot library depends on other libraries which brings us closer to dependency hell
- It is not possible to have axis on all sides

I will consider the attempt of changing the plot engine as a fun exercise, but for the above mentioned disadvantages I believe it is better to stick with gnuplot.

godsic:
@JeroenMulkers I still think it would be a good idea to have go-native visualization, assuming the backend is properly maintained. `gonum` appears to be a good choice, but [go-plotly](https://github.com/MetalBlueberry/go-plotly) might be a good alternative assuming we are using `offline` mode to avoid data breaches to 3rd parties.
In any case, @JeroenMulkers would you be so kind to split https://github.com/mumax/3/pull/272/commits/24c775e74decc846f3311bc76c7fa7e1d3147542 into self-contained ones, so that, it would be easier to revive this feature?

PR #283, opened 2021-02-12
Fix shifted quantity for shifts in more than one direction

JeroenMulkers:
The Shifted quantity could not be used to shift quantities over more than one direction at once. This PR fixes this issue.


godsic:
@JeroenMulkers Are unit tests possible for this one?

JeroenMulkers:
@godsic , I wrote a test in GO because it was too cumbersome to write a decent test in a mumax3 input script.

The test fails without the bug fix from this PR.

JLeliaert (approved):

Issue #298, opened 2021-11-09
error compiling mumax3


MKe2013:
Hi,

I have tried to compile mumax3 on Ubuntu. I have followed the installation instructions of "https://github.com/mumax/3", but I have not been lucky:

`cgo-gcc-prolog: In function ‘_cgo_c21ce16ee817_Cfunc_cuDeviceComputeCapability’:
cgo-gcc-prolog:56:2: warning: ‘cuDeviceComputeCapability’ is deprecated [-Wdeprecated-declarations]
In file included from /home/mike/go/pkg/mod/github.com/mumax/3@v3.9.3+incompatible/cuda/cu/device.go:5:
/usr/local/cuda/include/cuda.h:4195:36: note: declared here
 4195 | __CUDA_DEPRECATED CUresult CUDAAPI cuDeviceComputeCapability(int *major, int *minor, CUdevice dev);
      |                                    ^~~~~~~~~~~~~~~~~~~~~~~~~
cgo-gcc-prolog: In function ‘_cgo_c21ce16ee817_Cfunc_cuDeviceGetProperties’:
cgo-gcc-prolog:159:2: warning: ‘cuDeviceGetProperties’ is deprecated [-Wdeprecated-declarations]
In file included from /home/mike/go/pkg/mod/github.com/mumax/3@v3.9.3+incompatible/cuda/cu/device.go:5:
/usr/local/cuda/include/cuda.h:4161:36: note: declared here
 4161 | __CUDA_DEPRECATED CUresult CUDAAPI cuDeviceGetProperties(CUdevprop *prop, CUdevice dev);
      |                                    ^~~~~~~~~~~~~~~~~~~~~
# github.com/mumax/3/cuda/curand
/home/mike/go/pkg/mod/github.com/mumax/3@v3.9.3+incompatible/cuda/curand/generator.go:46:102: identifier "_Ctype_ulonglong" may conflict with identifiers generated by cgo
# github.com/mumax/3/cuda/cufft
/home/mike/go/pkg/mod/github.com/mumax/3@v3.9.3+incompatible/cuda/cufft/mode.go:17:52: could not determine kind of name for C.CUFFT_COMPATIBILITY_FFTW_ALL
/home/mike/go/pkg/mod/github.com/mumax/3@v3.9.3+incompatible/cuda/cufft/mode.go:16:52: could not determine kind of name for C.CUFFT_COMPATIBILITY_FFTW_ASYMMETRIC
[...]

PR #311, opened 2022-08-12
New build system migration (issue 298, and 2 more improvements)

Artemkth:
I have done several changes to original repo since I have forked, the most major of which is move to the new [go 'module' system.](https://go.dev/blog/using-go-modules) which allows building the Mumax3. Using old legacy arrangements for whatever reason selects wrong tag in git repo(v3.9.3), and produces no executables (highlighted in issues section https://github.com/mumax/3/issues/298).

Implementing proper build system is trivial for the project as there is no significant external dependencies, and just involves appending v3 to all the mumax3 imported modules per the new convention. And it also allows easier forking, as building is now possible outside of `$GOPATH`. Check out my personal fork repo readme to how it is done.

I also have added few more features of my own over the years, I didn't knew how to split the pull request so please consider those as well:
1. *New feature* I added the ability to use custom geometry from external geometry files with function `InitGeomFromOVF` which takes an external file name as an argument. It initialises both the mesh parameters, as well as sets the internal geometry to be the one from the external file. This allows to bake profiles externally in, for example, Wolfram Mathematica or Python. Example simulation, [Permalloy pig head](https://www.dropbox.com/s/05kko6armlh97f5/pig-head.png?dl=0), r~1um. The code for baking profiles I will release later. Geometry feature I extensively tested over last year, and haven't found any more bugs, yet, than I fixed in this PR.
2. Fixed OVF backend to follow the OVF standard specification more closely by allowing for both the whole line, and in-line comments. I've done verification with OOMMF checking that it accepts the comments. Mumax3 in main repo will crash when parsing a comment.

This version compiles for me, although, some minor changes to your CI scripts might be needed. Self tests run well, except the convergence test having a delta ~2e-5 higher than 1e-5 target.


JonathanMaes:
Thank you for your contribution! The two improvements have been merged in dc06ce32884b4fc18e8ca2203e493db3a0b9a378. The function `InitGeomFromOVF` was changed to an extension (it is now called `ext_InitGeomFromOVF`) because its functionality is somewhat more complex than normal functions: it does not only load a shape, but also changes the grid size. The `VoxelShape` is also no longer exposed to the user, because arrays are not supported in `.mx3` input scripts. This may change in the future.

The change to a Go module was not yet merged. What is the benefit of putting `/v3` in every import statement? I have created a `go.mod` file for the new mumax3.11 release, which compiles fine without adding `/v3`.

PR #316, opened 2022-11-07
Feature/fudgeShift

jsampaio:
A cleaned-up version of pull request #221.

This change introduces a new mode for the shift the problem window. In this new mode, the values of m in the introduced cells are copied from the cells at the border, instead of the values of ShiftMagL/R (current mode). This new mode is activated by setting FudgeShift = true (false by default).

When using auto-shifting procedures (e.g. ext_centerWall), it is still necessary to set the parameters ShiftMagL/R because they're used to determine the shift direction (as is the previous behaviour).

Modifications:
- In engine/shift.go: Added the user-accessible flag ‘FudgeShift’ to activate this shift mode. Modified engine.Shift.
- In the (cuda) implementation of the shift: new functions: cuda.ShiftFudgeX, cuda.ShiftFudgeY (and related .cu).
- added test/fudgeShift.mx3 to test to check if it behaves as announced.


JonathanMaes:
Thank you for your contribution, and apologies for the late response to your pull request.
Is there a reason why the fudgeShift is only implemented for the X and Y directions, but not Z?

JonathanMaes:
We decided to rename the boolean to `EdgeCarryShift`.
I understand now why you did not implement the Z-shift; only a shift along X is available in the input scripts (#337).

Issue #93, opened 2017-01-27
Crash when using MFM with mesh with Z size 2, 3, 4 or 5


jsampaio:
When the number of layers (mesh z size) is 2 or 4, mumax crashes when calculating the MFM kernel.  The error message comes from line 38 of mfmkernel.go :
```
36 util.AssertMsg(size[X]%2 == 0 && size[Y]%2 == 0, "Even kernel size needed")
37		if size[Z] > 1 {
38			util.AssertMsg(size[Z]%2 == 0, "Even kernel size needed")
39		}
```
Surprisingly (for me...), if you set the mesh size in X or Y to an odd number (127) it calculates the MFM kernel with no problems.

Here's the log of an interactive session, where I set the mesh Z size to 2 and selected the MFM image:

```
SetMesh(128, 64, 2, 4e-09, 4e-09, 4e-09, 0, 0, 0)
//resizing...
// Initializing geometry 0 %
// Initializing geometry 100 %
calculating MFM kernel
Even kernel size needed
panic: Even kernel size needed
[...]
	/home/joao/go/src/github.com/mumax/3/mag/mfmkernel.go:38 +0x631
github.com/mumax/3/cuda.NewMFM(0x1117e20, 0x3e6ad7f29abcaf48, 0x3f50624dd2f1a9fc, 0xc82115f880)
[...]
```

jsampaio:
This bug still happens, I can't follow the MFM code to correct it... if anyone wants to try, here's a mx3 script that triggers it:

```
Msat = 1e6; Aex = 10e-12;
nZ := 2;
SetMesh(128, 128, nZ, 4e-09, 4e-09, 4e-09, 0, 0, 0);
m  = vortex(1, 1);
print("Trying the MFM kernel for nZ=",nZ);
MFMLift = 40e-9; //forces the calculation of the MFM kernel.
print("survived nZ=",nZ);
```

this works for nZ=1, but crashes for nZ=2, 3, 4 or 5.

JonathanMaes:
This crash can be avoided by commenting out `mag/mfmkernel.go:113`:
```go
util.AssertMsg(size[Z]%2 == 0, "Even kernel size needed")
```
**The question is why this assert exists in the first place?** Without this assert, all tests still run as expected, and the resulting MFM images (generated by adding `Snapshot(MFM)` at the end of the previous comment's example script) seem just fine for `nZ <= 5`.

[...]

PR #334, opened 2024-10-03
Allow Nz=2,3,4,5 for MFM images

JonathanMaes:
This PR addresses issue #93.

It is unclear why the culprit, the now removed `AssertMsg` which checks if `Nz` is even, exists. Without it, all tests still run fine and MFM images still look as expected for the affected values of Nz.

There is another similar `AssertMsg` checking if `Nx` and `Ny` are even, but also for those everything seems to work as expected if the assert is deactivated. This is not yet included in the PR. Would it be safe to remove this too?

### DIFF
--- a/mag/mfmkernel.go
+++ b/mag/mfmkernel.go
@@ -3,11 +3,12 @@ package mag
 import (
 	"bufio"
 	"fmt"
+	"math"
+	"os"
+
 	d "github.com/mumax/3/data"
 	"github.com/mumax/3/oommf"
 	"github.com/mumax/3/util"
-	"math"
-	"os"
 )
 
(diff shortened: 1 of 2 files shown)


JonathanMaes:
Checked with odd Nx, Ny and all Nz<=10. Tested with a binary compiled on CUDA v12.6 for CC=8.6, and on CUDA v10.0 (the lowest we will support from 3.11 onwards) for CC=6.1. Both give identical and expected results. If the reason for the even kernel size requirement was CUDA-related, it seems to have been addressed before v10.0.

Therefore, I will merge this change, if anyone still gets errors related to this we can always rollback.

PR #335, opened 2024-10-03
Feature/complete grains at shape edge

JLeliaert:
Adds a variable to ext_make3dgrains to complete all voronoi grains whose centre lies within the shape. This also cuts away parts of the shape whose closest voronoi centre lies outside the shape.

This allows e.g. to make polycrystalline geometries consisting of clustered individual grains.

clusterparticle

### DIFF
--- a/engine/ext_make3dgrains.go
+++ b/engine/ext_make3dgrains.go
@@ -7,8 +7,11 @@ import (
 	"math/rand"
 )
 
+var CompleteGrainsAtShapeEdge  = false // complete the voronoi grains  instead of cutting them off at the shape edge
+
 func init() {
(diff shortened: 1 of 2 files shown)


JLeliaert:
One of the functions (shufflecells) in make3dgrains used rand.perm to shuffle the cells around before assigning them a region number. This function used the global random generator, which as of go 1.20 is seeded automatically and therefore gives a different output for every run. I've replaced this with the seeded generator of the tessellation to have a deterministic output. I've adapted the test problem to reflect the newly obtained values (the previous test values were obtained with a go version <1.20 which was still deterministic)

JonathanMaes (approved):

PR #341, opened 2024-10-17
In RK56: remove unnecessary torque evaluation

JonathanMaes:
In the RK56 solver, the torque was evaluated after the 9th stage. However, this torque was only used in `setMaxTorque()` a bit further in that function. By removing this unnecessary torque evaluation, the RK56 solver is >10% faster.

The following was not yet implemented, but is related: an additional performance gain could be achieved by not calling `setMaxTorque()` at all. It is not the cheapest function to calculate, and from a quick test it seems that not calling it would result in another ~5% improvement. The reason I say `setMaxTorque()` could be omitted, is because it just sets `LastTorque`, and it seems that that parameter is only used to show the maxTorque in the GUI, as well as being used in the Euler solver to determine its adaptive time step. The calculation of the maximum torque for `RelaxTorqueThreshold` does not use `LastTorque` at all, so this does not add to the utility of `setMaxTorque`. Whether `setMaxTorque()` could be removed is therefore up for debate.

### DIFF
--- a/engine/rk56.go
+++ b/engine/rk56.go
@@ -1,10 +1,11 @@
 package engine
 
 import (
+	"math"
+
 	"github.com/mumax/3/cuda"
 	"github.com/mumax/3/data"
 	"github.com/mumax/3/util"
-	"math"
 )
 
 type RK56 struct {
@@ -34,7 +35,6 @@ func (rk *RK56) Step() {
 	defer cuda.Recycle(k6)
 	defer cuda.Recycle(k7)
 	defer cuda.Recycle(k8)
-	//k2 will be recyled as k9
 
 	h := float32(Dt_si * GammaLL) // internal time step = Dt * gammaLL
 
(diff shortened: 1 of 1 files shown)


jplauzie:
Would this effect if you want to use Maxtorque as a condition in Runwhile()? For instance `Runwhile( MaxTorque >1e-4)` as in the [standard problem 4](https://github.com/mumax/3/blob/48fc687bfc35decb3bb1aaacdc7d0e72bc8a68ea/test/standardproblem4.txt) implementation?

JonathanMaes:
Thanks for your comment as always, I didn't think of that related function because apparently it is completely separate in the source code.

The `MaxTorque` that is exposed to the user in .mx3 scripts is calculated on-the-fly the moment it is reached in the input script. It's actually an alias for the internal function `GetMaxTorque()`, which determines the maximum of the `Torque` vector field (which is also exposed to the user, and which is similarly calculated at runtime whenever required).

I sense some optimization might be possible here, either by calculating the GUI maxTorque on-the-fly, or the other way around by making `MaxTorque` use the `LastTorque`. I suspect the latter could break some code, so that might not be the way to go.

PR #360, opened 2025-03-14
Feature/ext_enableCenterBubbleXY

ilateur:
Add enablers to center the bubble along X or Y direction during ext_centerBubble.


ilateur:
This is done again in #361 separately from #356.

Issue #352, opened 2025-01-28
bulk DMI is wrong/inconsistent for Neumann boundaries in 1D


DiegoDeGusem:
Hi,

When you create a 1D magnet along the x-axis with bulk DMI, mumax³ completely ignores the z-component by not calculating derivatives in the z-direction. I think it is wrong to ignore these derivatives, because when you compare this result with a magnet along the z-direction they are quite different.

Below you can find an input scrips for both cases, together with a plot of the final magnetization.

__x-direction__
```go
setcellsize(1e-9, 1e-9, 1e-9)
setgridsize(100, 1, 1)
msat = 0.86e6
aex = 13e-12
Ku1 = 0.4e6
anisU = vector(0,0,1)
Dbulk = 3e-3
enabledemag = False
openBC = False

m = Uniform(0,0,1)
minimize()
```

__z-direction__
```go
setcellsize(1e-9, 1e-9, 1e-9)
setgridsize(1, 1, 100)
msat = 0.86e6
aex = 13e-12
Ku1 = 0.4e6
anisU = vector(1,0,0)
Dbulk = 3e-3
enabledemag = False
openBC = False

m = Uniform(1,0,0)
minimize()
```

![Image](https://github.com/user-attachments/assets/60930501-ec76-4e77-822a-0f689d94093d)

ilateur:
This inconsistency is caused by the if-statement at [line 150 in dmibulk.cu](https://github.com/mumax/3/blame/master/cuda/dmibulk.cu#L150). For open BC this seems fine but not for Neumann BC.

JonathanMaes:
Addressed in the now merged pull request #363.

PR #363, opened 2025-03-26
Fix DMI Neumann boundary conditions for non-3D simulations

ilateur:
This should fix issue #352. Calculation of the effective magnetic field of bulk DMI was skipped when there was no derivative to be taken in the z-direction, but the Neumann boundary conditions do not cancel out. Open BC can still be skipped, and for interfacial DMI this is not a problem.
I also added a test to confirm the expected symmetry.

### DIFF
--- a/test/dmibulksymm1D.mx3
+++ b/test/dmibulksymm1D.mx3
@@ -1,7 +1,7 @@
 /*
-	Symmetry test for bulk DMI in a 1D magnet with Neumann BC.
-	Rotating the entire system by 90° should give symmetric result.
-    This test is based on the proposed 1D standard problem of bulk BMI:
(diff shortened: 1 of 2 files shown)


test/dmibulksymm1D.mx3
```diff
@@ -0,0 +1,43 @@
+/*
+	Symmetry test for bulk DMI in a 1D magnet with Neumann BC.
+	Rotating the entire system by 90° should give symmetric result.
+    This test is based on the proposed 1D standard problem of bulk BMI:
```
JonathanMaes, line 4: BMI typo?

ilateur: *DMI of course

PR #383, opened 2026-03-13
New shapes: Line and Line2D

JonathanMaes:
As suggested by @Mat1629, this PR adds two new shapes to create lines. This can be convenient for, e.g., defining 3D ASI structures.

**`Line(x1, y1, z1, x2, y2, z2, diam, linecap)`** generates a line from $(x_1, y_1, z_1)$ to $(x_2, y_2, z_2)$ with a diameter `diam`. The behavior of the line at these two endpoints is determined by the last argument `linecap`, which can be
- `"infinite"`: the line extends indefinitely beyond the two specified points.
- `"round"`: a sphere of radius `diam` is placed at the two endpoints, yielding a smooth shape.
- `"flat"`: the line stops abruptly at the two endpoints in a flat plane perpendicular to the line.

**`Line2D(x1, y1, x2, y2, diam, linecap)`** behaves the same, but extends indefinitely along the Z-axis. Hence, this basically creates a plane in 3D space whose normal vector always lies in the XY-plane.

A special case was implemented for `diam=0`, in which case the shape will be a minimally connected geometry of all cells intersected by the line.
> [!WARNING]
> If a `diam=0` line is scaled or rotated afterwards, it is no longer guaranteed that this shape is fully connected, since the shape function then has no way of knowing what cells it would eventually intersect.

## Visual examples

Below are some geometries from `test/line.mx3` (but after an additional `relax()`).

|  | `Line` | `Line2D` |
| --- | --- | --- |
| `"round"` | image | image |
| `"flat"` | image | image |
| `"infinite"`  | image | image|
 | `"infinite"` with `diam=0` | image | image |

### DIFF
--- a/engine/shape.go
+++ b/engine/shape.go
@@ -159,7 +159,7 @@ func Line2D(x1, y1, x2, y2, diam float64, linecap string) Shape {
 		return func(x, y, z float64) bool {
 			return diam*diam >= math.Pow((x-x1)*(y-y2)-(x-x2)*(y-y1), 2)/denom
 		}
-	case "round":
+	case "round", "flat":
 		return func(x, y, z float64) bool {
 			a, b := x2-x1, y2-y1
 			lenSq := a*a + b*b
@@ -167,6 +167,10 @@ func Line2D(x1, y1, x2, y2, diam float64, linecap string) Shape {
 			param := -1.0
 			if lenSq != 0 {
 				param = ((x-x1)*a + (y-y1)*b) / lenSq
+			}
+
+			if linecap == "flat" && (param < 0 || param > 1) { // If param is not in [0,1], then point is beyond line segment
+				return false
 			}
 
 			xx, yy := 0., 0.
@@ -202,7 +206,7 @@ func Line(x1, y1, z1, x2, y2, z2, diam float64, linecap string) Shape {
 			cross1, cross2, cross3 := dy1*dz2-dy2*dz1, dx1*dz2-dx2*dz1, dx1*dy2-dx2*dy1
 			return diam*diam >= (cross1*cross1+cross2*cross2+cross3*cross3)/denom
 		}
-	case "round":
+	case "round", "flat":
 		a, b, c := x2-x1, y2-y1, z2-z1
 		lenSq := a*a + b*b + c*c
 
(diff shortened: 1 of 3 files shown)


JLeliaert (changes requested):
The default value for the line cap is 0, which returns an error message
`type mismatch: can not use type int as string `

JLeliaert:
> A special case was implemented for diam=0, in which case the shape will be a minimally connected geometry of all cells intersected by the line.

When the line goes exactly through the corner, single precision leads to situations where a cell that should be selected is not, so that the geometry is no longer a face-connected set but is missing a "bridge" cell between two corner-connected cells.

see this example in 2D with a line along the diagonal
image

Adding a small buffer fixes this
`if tmax < tmin-1e-15 {`
image

ilateur (approved):

JLeliaert (approved):

engine/shape.go
```diff
@@ -137,6 +143,141 @@ func Triangle(x0, y0, x1, y1, x2, y2 float64) Shape {
...
+	case "infinite":
+		dx, dy := x2-x1, y2-y1
+		denom := dx*dx + dy*dy
+		return func(x, y, z float64) bool {
+			return diam*diam >= math.Pow((x-x1)*(y-y2)-(x-x2)*(y-y1), 2)/denom
+		}
+	case "round", "flat":
+		return func(x, y, z float64) bool {
+			a, b := x2-x1, y2-y1
+			lenSq := a*a + b*b
```
ilateur, line 166: This is called `dx`, `dy` and `denom` in the previous case. Can these be unified? They also do not depend on position and don't have to be recalculated at every position.

engine/shape.go
```diff
@@ -137,6 +143,141 @@ func Triangle(x0, y0, x1, y1, x2, y2 float64) Shape {
...
+	if diam == 0 { // Special case: shape consists of all cells that line intersects
+		return LineIntersectsCell([3]float64{x1, y1, 0}, [3]float64{x2, y2, 0}, 2, linecap)
+	}
+
+	switch linecap {
+	case "infinite":
+		dx, dy := x2-x1, y2-y1
+		denom := dx*dx + dy*dy
+		return func(x, y, z float64) bool {
+			return diam*diam >= math.Pow((x-x1)*(y-y2)-(x-x2)*(y-y1), 2)/denom
```
ilateur, line 161: Wait I think `diam` has the role of radius here

engine/shape.go
```diff
@@ -137,6 +143,141 @@ func Triangle(x0, y0, x1, y1, x2, y2 float64) Shape {
...
+			xx, yy := 0., 0.
+			if param < 0 {
+				xx, yy = x1, y1
+			} else if param > 1 {
+				xx, yy = x2, y2
+			} else {
+				xx, yy = x1+param*a, y1+param*b
+			}
+			dx, dy := x-xx, y-yy
+			return math.Sqrt(dx*dx+dy*dy) <= diam
```
ilateur, line 186: Wait I think `diam` has the role of radius here

engine/shape.go
ilateur, line 1: My comments apply to both 2D and 3D

ilateur (changes requested):

PR #384, opened 2026-03-18
Feature: Custom Quantities

jplauzie:
This branch adds the ability for the user to promote arbitrary slices (which can be made using NewScalarMask and NewVectorMask) into a custom Quantity. The implementation is backported from @XuebingZhao 's [umagnus](https://github.com/seeder-research/umagnus) fork. Currently, custom fields can only be made from existing Quantities, Const or Constvector , limiting the amount of spatially varying Quantities that are possible.

- The main use case is that it allows users to create spatially varying field terms that are not limited to 256 regions, but of course there are many other options. Does not use any Cuda, so probably slower than Jakelove's https://github.com/mumax/3/pull/289 implementation for spatial Quantities, but also probably more flexible overall.
- Works with both Scalar and Vector slices. Adding functionality to handle tensors isn't included and would be a bit of extra work, but users can fairly easily build a rank2 tensor out of Vectors.
- Works with normal Quantity operators (Add, Mul, etc), as well as with Run/Relax/Minimize etc.
- Included is a test file custom_quantities.mx3. I compared it to mumax's built-in uniaxial anisotropy, it matches exactly within numerical precision.
- In a previous PR, I mentioned potentially doing custom Excitations. After spending some more time on it, I believe it should actually be possible to do time varying Quantities as well, but it will take some more time to fully work it out and verify, so that will come at some later time in a different PR.


test/custom_quantities.mx3
```diff
@@ -0,0 +1,109 @@
...
+defregion(i,xrange(-(Nx*c/2)+i*c,-(Nx*c/2)+(i+1)*c).intersect(layer(0)))
+Ku1.setregion(i,1e6*(1-(i/Nx)))
+anisu.setregion(i, vector(1-1*(1-(i/Nx)), 1*(1-(i/Nx)), 0))
+}
+
+defregion(255,layer(1))
+Ku1.setregion(255,0)
+
+B_ext = vector(0, 0.10, 0)
+relax()
```
JonathanMaes: This test takes 30 seconds to run due to the use of `relax()`, but (if possible) tests should not take much longer than a few seconds. I suppose `run(1e-11)` would also be sufficient to check the correctness of the custom quantities? Then the test only takes 5 seconds (or even just 1 second with 1e-12).

engine/ext_customquantities.go
```diff
@@ -0,0 +1,42 @@
+package engine
+
+import (
+	"github.com/mumax/3/cuda"
+	"github.com/mumax/3/data"
+	"github.com/mumax/3/util"
+)
+
+func init() {
+	DeclFunc("CustomQuantity", CustomQuantity, "Creates a custom Quantity from the user provided (scalar or vector) slice.")
```
JonathanMaes: Should we mention `NewScalarMask` or `NewVectorMask` here as well to guide users to these otherwise obscure functions (in the sense that they are not prominently featured in the API)?

test/custom_quantities.mx3
```diff
@@ -0,0 +1,109 @@
...
+
+for i:=0; i<Nx; i++{
+Ku1.setregion(i,0.5e6*(1-(i/Nx)))
+}
+
+
+
+
+mask := newScalarMask(Nx, Ny, Nz)
+for i:=0; i<Nx; i++{
```
JonathanMaes, line 49: Perhaps having a CUDA kernel that generates a slice containing all the cell indices (similar to the functions from #289) could be useful for large simulations. In the example here, one could then use `.Comp(0)`, promote that to a quantity, and then operate on that quantity to efficiently build the mask.

But there were some unresolved comments on #289, so it's probably best to limit this PR to the slice-to-Quantity conversion and keep this idea for a potential future PR.

test/custom_quantities.mx3
```diff
@@ -0,0 +1,109 @@
...
+vecmask := newVectorMask(Nx, Ny, Nz)
+for i:=0; i<Nx; i++{
+	for j:=0; j<Ny; j++{
+		for k:=0; k<Nz; k++{
+		r := index2coord(i, j, k)
+		x := r.X()
+		y := r.Y()
+		z := r.Z()
+		
+		firstterm:=pow(  (1-1*(1-(i/Nx))),2)
```
JonathanMaes: Is there a reason to multiply by 1 here?

JonathanMaes (changes requested):

JonathanMaes:
I also think this feature is sufficiently useful to rename the file `ext_customquantities.go` to `customquantity.go`.

test/custom_quantities.mx3
jplauzie: Run(1e-12) is a better choice. Relax was convenient during initial testing to help verify it was working properly, but doesn't actually contribute to the test in a way that Run doesn't

engine/ext_customquantities.go
jplauzie: This makes sense. I also added a reminder that the slice should match the dimensions of the gridsize

edit: I also added a reminder that one can use Loadfile() as well. If something like .set() is too slow, a good way to get around it is to build the slice in something like Python and just Loadfile it in.

test/custom_quantities.mx3
jplauzie: The multiplication can be removed, it's just left over from earlier testing on whether the vectors were going to be automatically normalized

engine/ext_customquantities.go
jplauzie: I also added an extra reminder that it is not limited by the 256 region limit, since that's very unusual. The test restricts itself to 256 only in order to compare to the existing regions implementation, but the feature itself works fine with more.

test/custom_quantities.mx3
jplauzie: Something like this would be very useful at some point, especially because using .set() and .get() can end up being somewhat slow due to the extra overhead, for large simulations. Set isn't *too* bad by itself, but if a user wants to do something like Region.getcell to compare to a value, then .set() it gets really bogged down

JonathanMaes: I added a `CellIndices()` function inspired by #289 that returns a vector-`*data.Slice` containing the (x, y, z)-index of each cell. This avoids the moving window concern raised in #289, as the index of a cell remains constant at all times.
This can then of course be turned into a quantity using your `CustomQuantity(...)` function. Let me know if you have any remarks or further suggestions in this regard.

jplauzie: Nice work! It works well, I don't see any further necessary changes. It even works flawlessly with `.Comp()` and then promoting it to a Quantity

PR #385, opened 2026-03-18
Feature: RegionShapes

jplauzie:
This PR adds a function `AllRegionShapes` which for a given region value, will return a Shape with all cells in that region. Can be easily looped over all 256 regions. This is intended mainly if one wants to make Shapes of say, Voronoi grains, to be used with Masked().

- An example file is including showing how each shape can be accessed after AllRegionShapes is called (there is some slightly unusual syntax, it uses 2 sets of parantheses), and also how they can be added together in a simple loop, is attached here: [regionshape_example.txt](https://github.com/user-attachments/files/26082150/regionshape_example.txt)

- does not yet work with shift/moving frame stuff. I am not sure if it could, or should?

### DIFF
--- a/engine/ext_regionshapes.go
+++ b/engine/ext_regionshapes.go
@@ -1,0 +1,34 @@
+package engine
+
+import (
+	"github.com/mumax/3/data"
+)
+
+func init() {
+	DeclFunc("AllRegionShapes", AllRegionShapes, "Returns a function that gives the shape of each region.")
+}
+
+func AllRegionShapes() func(int) Shape {
+	mesh := Mesh()
+	arr := regions.HostList()
+	n := mesh.Size()
+	d := mesh.CellSize()
+	Lx, Ly, Lz := float64(n[X])*d[X], float64(n[Y])*d[Y], float64(n[Z])*d[Z]
+
+	return func(region int) Shape {
+		defRegionId(region)
+
+		return func(x, y, z float64) bool {
+			ix := int((x + 0.5*Lx) / d[X])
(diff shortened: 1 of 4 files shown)


engine/ext_regionshapes.go
```diff
@@ -0,0 +1,34 @@
+package engine
+
+import (
+	"github.com/mumax/3/data"
+)
+
+func init() {
+	DeclFunc("AllRegionShapes", AllRegionShapes, "Returns a function that gives the shape of each region.")
```
JonathanMaes: I think in the current state, where this feature doesn't work with shifting, this should probably be considered an extension (i.e., exposed as `ext_AllRegionShapes`) since we can't guarantee expected behavior in all cases.

This could be fixed if the Shape would adapt to changes in the regions array, but such mutability might not be desirable since it wouldn't allow one to save an old region configuration.

A short description of the possible caveats could be included in the description here to warn users.

engine/ext_regionshapes.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func AllRegionShapes() func(int) Shape {
```
JonathanMaes: The double parentheses syntax due to returning a `func(int)` could indeed be confusing, but I think this is ok for performance reasons and is not unprecedented (`FunctionFromDataFile` does so too).
The sentence in the API documentation is already pretty clear, but I would still include an adaptation of your example in the test directory to make its usage absolutely clear to users.

JonathanMaes (changes requested):

JonathanMaes (approved):

PR #386, opened 2026-03-18
Feature: GeometryEdgeFinder

jplauzie:
This PR adds a helper function that makes edges of the geometry (1 cell thick) into Shapes, allowing the user to easily define things for e.g. custom fields boundary conditions. By default, there are 6 Shapes corresponding to: +/- X, Y, Z edges.

- 2s permutations (corners such as +/- XY, etc) can be done by using existing Shape method .intersect(). 3s as well.
- Treats the edge of the simulation box as geometry edge. Does **not** currently check if PBC are enabled.
- example file attached: [geomedgefinder_example.txt](https://github.com/user-attachments/files/26082803/geomedgefinder_example.txt)

### DIFF
--- a/engine/ext_geomedgefinder.go
+++ b/engine/ext_geomedgefinder.go
@@ -1,0 +1,128 @@
+package engine
+
+import (
+	"math"
+
+	"github.com/mumax/3/cuda"
+)
+
+func init() {
+	DeclFunc("GeometryEdgePlusX", GeometryEdgePlusX, "Returns the +X edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgeMinusX", GeometryEdgeMinusX, "Returns the -X edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgePlusY", GeometryEdgePlusY, "Returns the +Y edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgeMinusY", GeometryEdgeMinusY, "Returns the -Y edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgePlusZ", GeometryEdgePlusZ, "Returns the +Z edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgeMinusZ", GeometryEdgeMinusZ, "Returns the -Z edge of the geometry as a Shape")
+}
+
+func GeometryEdgePlusX() Shape {
+	plusX, _, _, _, _, _ := GeometryEdges()
+	return plusX
+}
+
+func GeometryEdgeMinusX() Shape {
+	_, minusX, _, _, _, _ := GeometryEdges()
+	return minusX
+}
+
+func GeometryEdgePlusY() Shape {
+	_, _, plusY, _, _, _ := GeometryEdges()
+	return plusY
+}
+
+func GeometryEdgeMinusY() Shape {
+	_, _, _, minusY, _, _ := GeometryEdges()
+	return minusY
+}
+
+func GeometryEdgePlusZ() Shape {
+	_, _, _, _, plusZ, _ := GeometryEdges()
+	return plusZ
+}
+
+func GeometryEdgeMinusZ() Shape {
+	_, _, _, _, _, minusZ := GeometryEdges()
+	return minusZ
+}
+
+func GeometryEdges() (plusX, minusX, plusY, minusY, plusZ, minusZ Shape) {
+
+	s, r := geometry.Slice()
+	if r {
+		defer cuda.Recycle(s)
+	}
+
+	arr3d := s.HostCopy().Scalars()
+
+	n := Mesh().Size()
+	Nx, Ny, Nz := n[X], n[Y], n[Z]
+
+	edgeMasks := make([][]bool, 6)
+	for i := range edgeMasks {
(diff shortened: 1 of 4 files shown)


engine/ext_geomedgefinder.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/mumax/3/cuda"
+)
+
+func init() {
+	DeclFunc("GeometryEdgePlusX", GeometryEdgePlusX, "Returns the +X edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgeMinusX", GeometryEdgeMinusX, "Returns the -X edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgePlusY", GeometryEdgePlusY, "Returns the +Y edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgeMinusY", GeometryEdgeMinusY, "Returns the -Y edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgePlusZ", GeometryEdgePlusZ, "Returns the +Z edge of the geometry as a Shape")
+	DeclFunc("GeometryEdgeMinusZ", GeometryEdgeMinusZ, "Returns the -Z edge of the geometry as a Shape")
```
JonathanMaes: Since these functions won't play nicely with mesh resizes, I would put `ext_` in front of their name to signal that flawless operation is not guaranteed in every situation.

engine/ext_geomedgefinder.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func GeometryEdgePlusX() Shape {
```
JonathanMaes: I'm not sure if we have to introduce all these different functions. Perhaps a single function `ext_GeomEdge(axis string)` might be cleaner, using a switch block where `axis` can be `"+X"`, `"-X"`, `"+Y"`, `"-Y"`, `"+Z"` or `"-Z"` (or lowercase)?

engine/ext_geomedgefinder.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return maskToShape(edgeMasks[0], Nx, Ny, Nz), maskToShape(edgeMasks[1], Nx, Ny, Nz),
+		maskToShape(edgeMasks[2], Nx, Ny, Nz), maskToShape(edgeMasks[3], Nx, Ny, Nz),
+		maskToShape(edgeMasks[4], Nx, Ny, Nz), maskToShape(edgeMasks[5], Nx, Ny, Nz)
+}
+
+// Helper function to wrap bool mask into a Shape
+func maskToShape(mask []bool, Nx, Ny, Nz int) Shape {
+	return func(x, y, z float64) bool {
+		d := Mesh().CellSize()
```
JonathanMaes: Depending on the functionality we want, this line should either be inside or in front of the `func(x, y, z)`.
- On the one hand, since the geometry's extent in the mesh will change when the mesh changes, perhaps it makes more sense to put this before the function such that this edge shape scales in the same way?
- On the other hand, scaling this edge shape can never guarantee that it still represents the real edge of the geometry, so it might be better to keep this code as is and just warn people of the restrictions (doesn't change when the mesh/geometry changes) when using this function.

engine/ext_geomedgefinder.go
```diff
@@ -0,0 +1,128 @@
...
+	_, _, _, _, plusZ, _ := GeometryEdges()
+	return plusZ
+}
+
+func GeometryEdgeMinusZ() Shape {
+	_, _, _, _, _, minusZ := GeometryEdges()
+	return minusZ
+}
+
+func GeometryEdges() (plusX, minusX, plusY, minusY, plusZ, minusZ Shape) {
```
JonathanMaes: If this function isn't exposed to the users, I don't think it makes much sense to have it generate all the 6 edges since only 1 gets saved in the wrapping functions above. I would either expose it to the user for efficiency or just combine it with my comment above regarding `ext_GeomEdge(axis string)`.

JonathanMaes (changes requested):

engine/ext_geomedgefinder.go
jplauzie: Either way seems ok to me. I guess it doesn't hurt to try to have it scale?

engine/ext_geomedgefinder.go
```diff
@@ -44,6 +44,11 @@ func ext_GeometryEdge(axis string) Shape {
 		util.Log("GeometryEdge: invalid direction '" + axis + "'. Use +x, -x, +y, -y, +z, or -z")
 	}
 
+	PBCc := Mesh().PBC_code()
+	PBCx := (PBCc >> X) & 1
+	PBCy := (PBCc >> Y) & 1
+	PBCz := (PBCc >> Z) & 1
```
ilateur: 😣

JonathanMaes: I could replace this with `bool2float` from #392 if you prefer.

ilateur: yes(?)

JonathanMaes: sike! `bool2float` was not actually needed.

PR #387, opened 2026-03-18
Feature: Grain Boundaries

jplauzie:
This PR adds a new function that adds boundaries to existing regions, given user input of: (startregion, numgrains, offset, boundarythickness, zeroflag). Zeroflag: 1 = region0 is normal, 0 = region0 acts as edge but no boundary itself, -1 = ignore region0 entirely.

- Meant to be used after ext_makegrains or ext_make3dgrains, to give grains boundaries, but will of course work with any arbitrary shape. One can also assign all boundaries the same material parameters to have grains embedded into a single material, instead of each grain having individual boundaries.
- kind of region-intensive, since it will double the amount of regions used.
- can be called repeatedly if one wants multiple layers
- works in both 2D and 3D
- 1 major flaw: Since it works on the output of ext_makegrains, it currently cannot detect if you happen to have 2 grains that belong to the same region touching. There will be no boundary. The only way I can think to fix this is to have ext_grainboundaries call a modified ext_makegrains internally that uses only unique values but is unlimited to 256 regions, and then reduce (using modulo 256) back down to 256 after boundaries have been added (this is doable, but I have not implemented it yet, I figured I'd get feedback first). A bit unfortunate, since ideally you want to use a low number of regions to begin with since they get doubled.
- does not yet work with shift/moving frame stuff, where new grains will be introduced as the frame moves at the edges.

For instance, if one had called
ext_make3dgrains(200e-9, 1, 50, ellipsoid(5e-9*512,5e-9*256,5e-9*100), 12345678)

and you wanted 2 boundary layers each with thickness of 3 cells, you might do:
ext_grainboundaries(0,50,51,3,1)
ext_grainboundaries(0,50,102,3,1)


engine/ext_grainboundaries.go
```diff
@@ -0,0 +1,113 @@
...
+	size := r.Mesh().Size()
+	Nx, Ny, Nz := size[X], size[Y], size[Z]
+
+	host := r.HostList()
+
+	orig := make([]byte, len(host))
+	copy(orig, host)
+	origArr := reshapeBytes(orig, size)
+
+	if boundarythickness < 1 {
```
JonathanMaes: Would it be more convenient for users if boundarythickness were a length in meters, rather than an amount of cells?

engine/ext_grainboundaries.go
```diff
@@ -0,0 +1,113 @@
...
+					continue
+				}
+
+				if reg < startregion || reg > startregion+numgrains {
+					continue
+				}
+
+				isBoundary := false
+				for i := 0; i < len(dx); i++ {
+					nx := wrapPBC(ix+dx[i], X)
```
JonathanMaes, line 67: I made `wrapPBC` from the `ext_geomEdge` PR into a standalone function, since it can also be used here to easily account for PBC.

engine/ext_grainboundaries.go
```diff
@@ -0,0 +1,113 @@
...
+					ny := wrapPBC(iy+dy[i], Y)
+					nz := wrapPBC(iz+dz[i], Z)
+
+					if nx < 0 || nx >= Nx || ny < 0 || ny >= Ny || nz < 0 || nz >= Nz {
+						continue
```
JonathanMaes, line 80: I initially found it strange that the simulation edge is not a "boundary" (for open BC), but I understand this is necessary for Nz=1 systems which would otherwise be pure boundary.

I don't know if it would be worth adding more complexity to address this. If we do, a separate variable could be used to avoid even more function arguments. Some options include:
- An int whose last 3 bits control whether the Z-, Y- and X-edges are considered part of the boundary (Ian would probably respond "😣" to this suggestion).
- 3 boolean variables to achieve the same as the previous option.
- A single boolean variable (default: false) that controls whether the Z-edges are considered part of the boundary, while the X- and Y-edges are always considered part of the boundary (this is likely undesirable since this will lead to artifacts in geometries that come close to the edge).
- Leaving it as is, perhaps adding a small note to the documentation that the simulation edge is never considered a region boundary.

engine/ext_grainboundaries.go
jplauzie: That makes sense. This was originally converted over from a python script, but doing it in meters is probably more intuitive for users.

That does open up the question, how to handle if the cellsize is different in x/y/z? in particular for line 39. Right now, it's just an isotropic sphere regardless of cellsizes

engine/ext_grainboundaries.go
jplauzie: Oops. I originally had a flag for this in development, it must've got lost in a refactoring. But I didn't consider breaking it into separate flags for X/Y/Z, it was just on or off. Splitting them up makes more sense, so you can have boundaries in XY for a 2D simulation where Nz=1.

I think 3 booleans works? It seems like it would be hard to convey to a user how to manipulate the last 3 bits of an int.

engine/ext_grainboundaries.go
JonathanMaes: While writing a response, I had pretty much written the required code so I have pushed it in a commit (bdaaae9f). Feel free to correct any inefficiencies or mistakes in my code.

engine/ext_grainboundaries.go
JonathanMaes: 3 boolean variables is all right for me.

Would we then, as default,
- not treat edges as boundaries, or
- treat the X- and Y-edges as boundaries but not the Z-edge?

I wouldn't treat all edges as boundaries by default to avoid the aforementioned Nz=1 issue.

ilateur: 3 booleans also seems like the best option to me

jplauzie: Yeah, I would definitely default Z as off, it's kind of disorienting if you aren't expecting it. For X/Y, either default seems fine. I guess on by default would probably make more sense for a new user?

engine/ext_grainboundaries.go
jplauzie: This looks like it works well. It does mean if you do something like x-cellsize=4e-9, y-cellsize=8e-9, and bordersize=4e-9, you do get no boundaries on the y-boundaries. And vice versa, bumping bordersize up to 8e-9 will give you a thickness of 2 on the x-boundaries
grains

The alternative would be to force at least one cell? This isn't "broken"/wrong as is though, and in fact it's more consistent with how Shapes already work. There isn't a perfect answer, since it depends fundamentally it is trying to resolve a feature that is less than the cellsize, so either choice has trade-offs

[graintest.mx3.txt](https://github.com/user-attachments/files/26622455/graintest.mx3.txt)

engine/ext_grainboundaries.go
ilateur: X and Y off would be the most symmetric which I like, but X and Y on is probably the most intuitive in most situations?

JonathanMaes: I agree, but there are plenty of functions in mumax³ that already break symmetry by making the Z-axis special. Since X and Y on indeed seems most intuitive, I would go for that as the default then.

engine/ext_grainboundaries.go
ilateur: I would say this is the user's responsibility, because this is consistent with shapes in general.

If you really want to be nice, you could add a warning if the given thickness is smaller than any (relevant) cell sizes... but then what if the mesh changes later again..?

JonathanMaes: Ah, I hadn't thought of that. Since this is expected behavior given a certain boundarythickness, I would consider it the user's responsibility to not rely on feature sizes smaller than a cell.

jplauzie: That works for me

jplauzie:
Added the 3 bools as ext_grainboundary_edgeX/Y/Z. They respect PBC, so if PBC is on in say X, but ext_grainboundary_edgeX is false, you'll still get boundaries as expected, if there are two different regions on either side of the wrap

PR #388, opened 2026-03-18
Feature: Wall clock timers for RunWhile, Relax, Minimize

jplauzie:
Adds a function that allows the user to add some noise, then Run/Relax/Minimize for a specified amount of wallclock time or (for Relax/Minimize) if the solver converges. This is useful for stiff problems where one particular solver gets stuck (In my experience, this can happen with e.g. hysteresis loops at small external fields), to help jiggle out of a local minimum.

- noise, Runwhile, Relax, Minimize , are each consecutively called for some user-input wallclock time. If a user wants a different order (say, no Run), they can just input 0 walltime, judiciously. This can also be looped in a while(for) loop to keep cycling between run->relax->minimize until it converges.
 - The current implementation is a bit messy, I basically copied Minimize/Relax verbatim. It would be cleaner to integrate the inputs/flags into existing Relax/Minimize directly (or possibly using injections?) but I didn't want to mess with such sensitive core functions.
- Introduces a new magnetization Config, CurrentMag. This just makes the current magnetization into a Config. This is mostly useful for adding noise, as "m=Currentmag().add(0.1,randommag())". Works with stuff like .transl() as one would expect. Can be a bit memory intensive as it stores a Hostcopy of the magnetization whereas Configs usually are a mathematical function. I have not seen any issues with e.g. memory leaks.


engine/ext_wallclocktimer.go
```diff
@@ -0,0 +1,264 @@
...
+
+func init() {
+	DeclFunc("MinimizeForSeconds", MinimizeForSeconds, "Minimize for a fixed wall-clock time (int seconds)")
+	DeclFunc("RelaxForSeconds", RelaxForSeconds, "Relax for a fixed wall-clock time (int seconds)")
+	DeclFunc("RunForSeconds", RunForSeconds, "Run the simulation for a fixed wall-clock time (int seconds)")
+	DeclFunc("CurrentMag", CurrentMag, "Returns the current magnetization as a Config. E.g. CurrentMag().Add(0.1, RandomMagSeed(123)) will return a Config with the current magnetization plus some noise.")
+	DeclFunc("RunSequence", RunSequence, " (total wallclock time, noise, wallclock1, wallclock 2, wallclock3, seed). Runs a sequence of steps: 1) Add noise to current magnetization as CurrentMag().Add(noise, RandomMagSeed(seed)), 2) Run for wallclock1 seconds, 3) Relax for wallclock2 seconds, 4) Minimize for wallclock3 seconds. The sequence is repeated until convergence or total wall-clock time limit is reached.")
+}
+
+func MinimizeForSeconds(seconds int) (converged bool) {
```
JonathanMaes: This function looks very similar to `engine.Minimize()`, apart from different comments and the 3 places where time is taken into account, so I think it would be beneficial to combine them to avoid duplicate code.

The original `engine.Minimize()` could, for instance, be replaced by a call to `MinimizeForSeconds((1<<63 - 1) / int(time.Second))` (an integer "infinity"), while `engine.MinimizeForSeconds(seconds int)` could then be put in `engine/minimizer.go`.
I don't think there's much reason to consider this an "extension" (ext_), since not much unintuitive behavior can occur from this apart from possible unintended side effects from interrupting the minimizer mid-minimization.

engine/ext_wallclocktimer.go
```diff
@@ -0,0 +1,264 @@
...
+	"math"
+	"time"
+
+	"github.com/mumax/3/cuda"
+	"github.com/mumax/3/data"
+	"github.com/mumax/3/util"
+)
+
+func init() {
+	DeclFunc("MinimizeForSeconds", MinimizeForSeconds, "Minimize for a fixed wall-clock time (int seconds)")
```
JonathanMaes: Perhaps a small warning could be added that interrupting the minimizer before convergence does not guarantee correct solutions. It would also be useful to tell the user that the boolean return value indicates whether convergence was reached.

engine/ext_wallclocktimer.go
```diff
@@ -0,0 +1,264 @@
...
+	DeclFunc("RelaxForSeconds", RelaxForSeconds, "Relax for a fixed wall-clock time (int seconds)")
```
JonathanMaes: Same comment, except that interrupting `relax()` should be fine at any point in time so no warning is needed here.

engine/ext_wallclocktimer.go
```diff
@@ -0,0 +1,264 @@
...
+	DeclFunc("RunForSeconds", RunForSeconds, "Run the simulation for a fixed wall-clock time (int seconds)")
+	DeclFunc("CurrentMag", CurrentMag, "Returns the current magnetization as a Config. E.g. CurrentMag().Add(0.1, RandomMagSeed(123)) will return a Config with the current magnetization plus some noise.")
+	DeclFunc("RunSequence", RunSequence, " (total wallclock time, noise, wallclock1, wallclock 2, wallclock3, seed). Runs a sequence of steps: 1) Add noise to current magnetization as CurrentMag().Add(noise, RandomMagSeed(seed)), 2) Run for wallclock1 seconds, 3) Relax for wallclock2 seconds, 4) Minimize for wallclock3 seconds. The sequence is repeated until convergence or total wall-clock time limit is reached.")
```
JonathanMaes: I think this function might be too specialized to include in the standard mumax³ library. It seems like this is something users can just implement in an `.mx3` script if they would need it, unless I'm missing something here?

engine/ext_wallclocktimer.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/mumax/3/cuda"
+	"github.com/mumax/3/data"
+	"github.com/mumax/3/util"
+)
+
+func init() {
+	DeclFunc("MinimizeForSeconds", MinimizeForSeconds, "Minimize for a fixed wall-clock time (int seconds)")
+	DeclFunc("RelaxForSeconds", RelaxForSeconds, "Relax for a fixed wall-clock time (int seconds)")
+	DeclFunc("RunForSeconds", RunForSeconds, "Run the simulation for a fixed wall-clock time (int seconds)")
+	DeclFunc("CurrentMag", CurrentMag, "Returns the current magnetization as a Config. E.g. CurrentMag().Add(0.1, RandomMagSeed(123)) will return a Config with the current magnetization plus some noise.")
```
JonathanMaes: Seems like a very useful function (it could already come in handy in #391), but I would put it in `engine/config.go`.

engine/ext_wallclocktimer.go
JonathanMaes: It seems like the `time.Duration` doesn't play nicely with floats, but I wonder if it would be worth making `seconds` a floating-point number. Though on the other hand, an integer amount of seconds is probably already sufficiently flexible.

engine/ext_wallclocktimer.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+	}
+
+	pause = true
+	converged = time.Since(start) <= time.Duration(seconds)*time.Second
+	stepper.Free()
+	return converged
+}
+
+func RunForSeconds(seconds int) {
```
JonathanMaes: Seems perfectly fine, can go below `Run()` in `engine/run.go`.

engine/ext_wallclocktimer.go
```diff
@@ -0,0 +1,264 @@
...
+
+	RunWhile(cond)
+	pause = true
+
+	converged = time.Since(start) <= time.Duration(seconds)*time.Second
+	stepper.Free()
+	return converged
+}
+
+func RelaxForSeconds(seconds int) (converged bool) {
```
JonathanMaes: Similar comment but for `engine.Relax()` in `engine/relax.go`.

JonathanMaes (changes requested):

engine/ext_wallclocktimer.go
ilateur: Is there no cleaner solution than "integer infinity"? perhaps by making the seconds optional (my Go is not good enough to know if this is possible)?

JonathanMaes: Sadly, Go has no optional arguments. I think this is the one thing that annoys me most in Go.
A cleaner solution would be to use float infinity, but that has the issue of making `time.Duration` overflow to a negative duration, so a check should then be added to consider a negative `seconds` to be equivalent to infinity.

jplauzie: Yeah, my original intent was to make it an optional argument, and I couldn't find a way.  Perhaps just put a little helper function that returns a bool? For any negative seconds it always returns true. For 0, it returns false. For seconds>0, it checks if the current time is greater than the given duration and returns true if less, and false if greater. Instead of exposing the helper function at all, it just gets called internally in minimizer, but the seconds variable can be exposed to the user and initialized as -1, similar to how variables like MaxErr work. That way, if the user doesn't change it, minimize just reverts to the old behavior, and it stays backwards compatible with old scripts.

Something like:

func minimizeTimeLimit(start time.Time, seconds float64) bool {
	if seconds < 0 {
		return true
	}
	if seconds == 0 {
		return false
	}
	return time.Since(start) < time.Duration(seconds)*time.Second
}

I think that's basically what you said above, but without the infinity

A quick proof of concept as minimizer_v2 [minimizer_v2.go.txt](https://github.com/user-attachments/files/26239484/minimizer_v2.go.txt) (sorry if there's a better way to upload a different variant, I'm not sure if github has a way to maintain 2 different variants of the file in the same PR). I threw it together quickly so I'd need to double check everything, but maybe something like that?

JonathanMaes: That seems like a nice and clean way to do it.
I indeed don't think a PR can contain 2 variants of the same file, but feel free to update the PR to this new version. 
Note that creating a `time.Duration(float64)` will round the float to an integer, since Go internally uses an integer number of nanoseconds to represent a duration. Perhaps `time.Duration(seconds*time.Second)` would work as intended, since `time.Second` is simply the number one billion.

engine/ext_wallclocktimer.go
jplauzie: You're right, this can be accomplished with an appropriate for-loop in an `.mx3` script. The Go function doesn't add anything

jplauzie:
This should be a bit cleaner implementation, I think.

- Moves CurrentMag() to config.go
- folds RelaxForSeconds and MinimizeForSeconds into Relax/Minimize respectively via a small WallTimer function, and exposes RelaxWallClockTime and MinimizeWallclockTime to the user as floats they can set
- RunSequence was removed, as this can be done in an mx3 file
- Since RunWhile already accepts a bool argument, RunForSeconds can also already be accomplished in an mx3 file, and RunForSeconds was effectively just a wrapper function, so I removed it to keep the code base a bit cleaner

engine/relax.go
```diff
@@ -4,23 +4,39 @@ package engine
...
+	DeclFunc("Relax", Relax, "Try to minimize the total energy. Returns true if convergence is reached, or false if the wall-clock time limit is exceeded. The wall-clock time limit is disabled by default.")
 	DeclVar("RelaxTorqueThreshold", &RelaxTorqueThreshold, "MaxTorque threshold for relax(). If set to -1 (default), relax() will stop when the average torque is steady or increasing.")
+	DeclVar("RelaxWallClockTime", &RelaxWallClockTime, "Wall-clock time limit (seconds) for Relax that will interrupt the relaxation if exceeded. Set to -1 (default) to disable.")
 }
 
 // are we relaxing?
 var relaxing = false
 
-func Relax() {
+func Relax() (RelaxConverged bool) {
```
JonathanMaes: This should be `func Relax() bool {`; the way it's written now declares `RelaxConverged` as a variable inside the `Relax()` function, unrelated to the `RelaxConverged` defined on line 17. It is done correctly in `Minimize()`.

engine/relax.go
JonathanMaes: This loop always runs fully, and can thereby cause the runtime of `relax()` to exceed `RelaxWallClockTime`. Such an issue is not present in `minimize()`, which nicely works even with non-integer second durations :)

JonathanMaes (changes requested):

JonathanMaes (approved):

test/wallclocktime.mx3
```diff
@@ -0,0 +1,32 @@
...
+
+//// Interrupt relax/minimize
+B_ext = Vector(0, 0, 0)
+t_range := 0.005
+
+m = RandomMag()
+RelaxWallclockTime = 0.1 // Takes 49s to converge
+t0 := now()
+ExpectB("Relax not converged", Relax(), false)
+Expect("Relax stopped in time", since(t0).Seconds(), RelaxWallclockTime + t_range, t_range)
```
ilateur: why `+ t_range`?

JonathanMaes: Such that the captured range is [RelaxWallclockTime, RelaxWallclocktime + 2*t_range]. Though I guess it should be t_range/2 to be clear.

ilateur: Ah like that, I would've naively expected something like [RelaxWallclockTime - t_range, RelaxWallclocktime + t_range], but I guess the first half of that range would never occur.

PR #389, opened 2026-03-18
Bugfix: fix typo for 255 check in regions.go

jplauzie:
I think there may be two typos in the max number of allowed regions?

In regions.Loadfile() and defRegionId(), it checks if regions are >256. However, I believe this should be >=256? Regions allows 256 regions (0-255), but it should not allow 256. 

There is a similar typo in the Voronoi example on the website example page: The line:
"defregion(256, circle(N*c).inverse())". 

The Setgeom() command in the example moots it, but 256 actually wraps back to region 0. (This PR does not fix this)


JonathanMaes:
Thanks for noticing this. In the meantime, I also exposed `NREGION` for use in .mx3 scripts and changed some usage of region 256 to region 0 in test scripts.

PR #390, opened 2026-03-18
Re-expose Voxelshape to user

jplauzie:
In 3.11, Voxelshape was restricted from users. However, while the description says 'array', it actually works with slices. In particular, it works with slices created with NewScalarMask, with no other changes needed (surprisingly, no changes were needed to manage GPUAccess/CPUAccess at all). This can probably be safely re-exposed to the user? A minimal working example is attached: 
[testvoxel.txt](https://github.com/user-attachments/files/26085155/testvoxel.txt)


JonathanMaes:
Seems alright to expose this to the user. I adapted your MWE into a small test script and slightly tweaked the description.

ilateur (approved):

PR #391, opened 2026-03-23
Bugfixes in freelayerthickness.mx3 test

jplauzie:
Hi,

When trying to answer this [discussion question](https://groups.google.com/g/mumax2/c/DUDdYtNgNfc/m/770FgnBZCgAJ), I ran into some strange behavior on the Slonczewski STT test [here](https://github.com/mumax/3/blob/a0c5126ac5477b02ee321bebfb2ed5e2d76368b0/test/freelayerthickness.mx3#L4). My understanding is, FreeLayerThickness(FLT) should be set as the full thickness of the FM (e.g., in the [mumax workshop](https://mumax.ugent.be/mumax3-workshop/example_scripts/session4_example3b.txt) for 3 layers of 1e-9, with 1 nonmagnetic spacer layer, they use FLT=2e-9). 

This test has a few issues:

Issue1: The test passes despite torques being different (you can print out the torques), due to tolerance of 1e-5 being too loose. I believe 1e-5 is absolute error instead of relative

Issue 2: FLT is defined as thickness of the FM. For explicit setting in Case2, this should be 2e-9, not 1e-9, in order to be comparable to Case1 where FLT is unset but the gridsize is 2 and cellsize is 1e-9.

Issue3: When you change the gridsize in Z from 2 to 4, this changes the magnetization of the cell. Set m explicitly to be consistent, otherwise the torques still won't match. I set it to what the original test used, but the m could be something simpler. After both setting m and FLT to 2e-9, they match.

Issue4: (**NOT** fixed in this PR): If you change the cellsize after SetGeom, STT torque is silently 0 and not set. If you do Setcellsize before setgeom it's fine.  Seems to be because the cell magnetization is (0,0,0). I am guessing it's just a geometry clipping issue because it happens if you increase the cellsize to 2e-9, but not if you decrease it to 0.5e-9. Kind of weird, because cell(0,0,0) should not be getting clipped out of the geometry or anything for a Setgeom(layers(0,2)).

Attached is a slightly modified test script that shows these issues in the original:
[sttbug4.txt](https://github.com/user-attachments/files/26172751/sttbug4.txt)

This PR has a fixed test that should fix Issues 1-3 by setting a more strict tolerance (I don't know what the normal value is for torques? They're on the order of 1e-7 for these parameters. I know there are numerical noise issues at some point), setting the FLT explicitly on line 34 to 2e-9, and setting m in both cases explicitly.

(Peeking at Mumax+, it has a similar [test ](https://github.com/mumax/plus/blob/master/test/test_free_layer_thickness.py#L54) , but the implementation seems slightly different. It uses relative error of 1e-5, and explicitly sets the FLT to 1e-9 (not 2e-9? seems like it should be 2e-9 for the same reason as here) for both cases). There are some (different) strange things in that test (torque_result uses magnet_wanted, and the torques seem to match even if i explicitly set different FLTs) so I will open a separate issue over there once I'm sure I fully understand it, if it's still weird. I'm not as good with Mumax+ yet.)

Best regards,
Josh L.


test/freelayerthickness.mx3
```diff
@@ -27,12 +31,14 @@ torqueWanted := Vector(
...
 
 // Computation of the Slonczewski torque in the bottom cell
 torqueResult:= Vector(
     STTorque.hostCopy().Get(0,0,0,0),
     STTorque.hostCopy().Get(1,0,0,0),
     STTorque.hostCopy().Get(2,0,0,0))
 
-expectv("sttorque",torqueResult,torqueWanted,1e-5 )
\ No newline at end of file
+expectv("sttorque",torqueResult,torqueWanted,1e-8 )
```
ilateur: As an absolute error this still seems awfully low, because the torques are of order 1e-8 as well.

jplauzie: Agreed. Do you know what would be a safe value? I don't know where numerical noise becomes an issue for torques, normally.

There are not many tests using torques, but test/fixedlayer.mx3 seems to use a TOL 1e-10. However, in the relax code, it seems to stop at only 1e-9? The test passes for me at 1e-10 or even smaller

ilateur: Numerical noise is often only relative, so at best about 1e-7 * value, but probably a bit larger. These are also not normal torques, because of  the arbitrary parameters (like `Msat = 4.3`), so it's even harder to say in absolute terms.

The point is just that they are *basically* the same, so dividing both torque vectors by the norm of one before using `expectV` should yield a good relative comparison (which I am quite fond of), but this seems to be a bit unusual for mumax3 tests

JonathanMaes: The result is actually exactly the same (as it should be):
```javascript
print(torqueResult)
//[6.640856753392654e-08 7.62618412863958e-09 1.0106999326353616e-07]
print(torqueWanted)
//[6.640856753392654e-08 7.62618412863958e-09 1.0106999326353616e-07]
```
so it's even safe to use 0 relative tolerance. Since mumax³ is supposed to be deterministic, this test failing would indicate a subtle difference was somehow introduced, warranting a closer look.

test/freelayerthickness.mx3
```diff
@@ -10,11 +10,15 @@ Msat = 4.3
 J = vector(0,0,5.4)
 Pol = 0.6
 FixedLayer = vector(1,0,1)
+print(m.getcell(0,0,0))
+//For consistency, we should keep m(0,0,0) the same between tests. It will change when we change the gridsize, otherwise
+m=uniform(0.8096410632133484, -0.29036206007003784, -0.5100698471069336)
```
JonathanMaes, line 15: We could also make use of the `CurrentMag()` function introduced in your other pull request to copy the initial random magnetisation of these layers into the new enlarged mesh.

ilateur: I don't see a problem with using numbers like this though. It just seems simpler

jplauzie: Either way seems ok to me

JonathanMaes:
I think the bug you describe in point 4) is expected behavior.

`Layers(a, b int)` is simply a wrapper around `ZRange(a, b float64)`, so when you change the cellsize, this geometry gets scaled like any other.

I agree that the name `Layers` suggests differently, and I don't think anything prevents a fix that makes it always represent a particular range of layers regardless of changes in mesh size. However, that would break backwards compatibility, and might be confusing in a different way since no other Shape depends on the mesh size like that. Perhaps a warning or clarification in its documentation would be more appropriate.

What are your thoughts on this @jplauzie @ilateur?

jplauzie:
Ahh, that actually makes perfect sense now, for 4). I tricked myself. I usually use m.setinshape to visualize a shape instead of setgeom, because it makes it easier to pinpoint mistakes in the GUI. But m gets stretched to fit the new universe when you change the mesh. So if you do something like m.setinshape(Layers(0,2),uniform(1,0,0)) and then change the cellsize, the (1,0,0) automatically adjusts to stay on the two top layers, even after the change. I was thinking it was Layers doing that, but it's actually just m readjusting. geom doesn't do that. Oops 😅

Knowing it wraps Zrange and mumax puts the origin at the center, it should indeed work like that. I think it makes sense to leave it as is, especially for backwards compatibility. A warning/clarification should be sufficient

PR #392, opened 2026-04-02
Add ExpectB and bool2float functions

JonathanMaes:
`ExpectB()` is the same as `Expect()` and `ExpectV()`, but for comparing boolean values.
This could, for instance, be useful to test #388, and the new internal function `bool2float` can also be used to replace the shift operation in #386.

### DIFF
--- a/engine/util.go
+++ b/engine/util.go
@@ -7,7 +7,6 @@ import (
 	"path"
 	"sort"
 	"strings"
-	"unsafe"
 
 	"github.com/mumax/3/cuda"
 	"github.com/mumax/3/data"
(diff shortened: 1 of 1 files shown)


engine/util.go
```diff
@@ -70,6 +72,11 @@ func ExpectV(msg string, have, want data.Vector, maxErr float64) {
 	}
 }
 
+func ExpectB(msg string, have, want bool) {
+	haveInt, wantInt := bool2float(have), bool2float(want) // Fast boolean conversion
+	Expect(msg, haveInt, wantInt, 0)                       // Reuse Expect() to reuse same logging
```
ilateur: this seems silly but okay sure

engine/util.go
```diff
@@ -178,6 +185,11 @@ func paramDiv(dst, a, b [][NREGION]float32) {
 	}
 }
 
+// Converts true to 1.0, false to 0.0
+func bool2float(b bool) float64 {
+	return float64(*(*byte)(unsafe.Pointer(&b)))
+}
```
ilateur: It's not even the fastest method 😭 
(and a simple if-statement is way more readable than unsafe pointer magic)

JonathanMaes: I didn't call it fastBool2float :P
But ok, I changed it to a more readable function which should allegedly be 3% faster.

engine/util.go
JonathanMaes: Oops I forgot I changed it to a float after first using an int.

ilateur (approved):
perfect

PR #393, opened 2026-04-16
(draft) Feature: Time quantity

jplauzie:
Hi,

This is a first stab at adding the time varying custom Quantities functionality I mentioned in https://github.com/mumax/3/pull/384. It's split into 3 PRs for easier review, but they're intended to be used together. Sorry for not lumping it in with the other changes, I expected this to take longer. This should be the last PR for a bit on mumax3, I know I've dumped a lot on you all at once. 😅 As always, feel free to decline/request changes/take your time reviewing etc. 

Combined with CustomQuantities https://github.com/mumax/3/pull/384 , this should allow a user to build Quantities that are both spatially and time dependent. Currently, mumax3 is limited to separable functions f(x,y,z) * g(t). This should allow for functions that can't be separated like e.g. a Gaussian wave packet e^((x-v*t)^2/ (2 * sigma)). This PR adds 2 things:

1) A wrapper that promotes the Time variable into a Quantity. This Quantity automatically updates as the Time changes. I actually found two different implementations: One uses the existing NewVectorField/NewScalarField functions, and one uses a little struct wrapper similar to CustomQuantities. As far as I can tell, they're essentially equivalent - Do you have a preference? 

2)  For convenience, there are 2 new functions ScalarFuncQ and VectorFuncQ. This allows the user to give a function input (e.g. "VectorFuncQ(sin(t), 0, cos(t))"). This is mostly redundant with the other functionality in these PRs, so I'm not sure if it's worth including. The 2 potential benefits:
- You can do something like 0.1*sin(w*t) in one shot and it saves you some kernel calls instead of having to do something like Mul(Const(0.1), Qsin( Mul(w,t))). In testing, it does seem to be noticeably faster than the equivalent individual kernel calls, but takes a hit if you add it to the table (because then it pulls the data back off the GPU). With the table overhead it was ~equivalent speed wise.
- Strictly speaking, just doing t is a valid function of g(t), so maybe 1) should just be a subset of this instead of being it's own Quantity?

Included is a test file. To compare, I tested it against mumax's built in B_ext.add(), comparing sin(k*x-w*t) to the (using trig identities, sin(kx-wt)=sin(kx)cos(wt)-cos(kx)sin(wt) ) separable version. It seems to match well. The test file requires Custom Quantities, this Time Quantity, and the unary math kernels in the 2nd PR to run.

### DIFF
--- a/engine/ext_timequantity.go
+++ b/engine/ext_timequantity.go
@@ -44,7 +44,7 @@ func (t *timeQuantity) EvalTo(dst *data.Slice) {
 	}
 }
 
-// functions of time as quantities:
+// functions of time as Quantities:
 
 type scalarFuncQuantity struct {
 	f script.ScalarFunction
--- a/engine/ext_timequantity.go
+++ b/engine/ext_timequantity.go
@@ -1,0 +1,77 @@
+package engine
+
+import (
+	"github.com/mumax/3/cuda"
+	"github.com/mumax/3/data"
+	"github.com/mumax/3/script"
+)
+
+func init() {
+	DeclROnly("timeQ_s", TimeQ_s, "Current simulation time as a scalar Quantity (s)")
+	DeclROnly("timeQ_v", TimeQ_v, "Current simulation time as a vector Quantity (s)")
+	DeclROnly("timeQ_s2", TimeQ_s2, "Current simulation time as a scalar Quantity (s)")
+	DeclROnly("timeQ_v2", TimeQ_v2, "Current simulation time as a vector Quantity (s)")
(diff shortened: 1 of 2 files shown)


JonathanMaes:
It would appear that my deletion of the `mumax:3.12` branch has inadvertently closed these draft PRs.
In the style of mumax+, I have now made a new `mumax:dev` branch that can be used to prepare any future release, and avoid this issue in the future.

jplauzie:
Ahh, sorry about that, I didn't mean to mess with the 3.12 release. I will reopen these on the dev branch

PR #401, opened 2026-06-14
Allows odd N mesh for nonzero Temp

jplauzie:
Hi,

This PR adds support for using odd grids with temperature enabled. Previously, N (N=Nx*Ny*Nz) was required to be even when temperature was nonzero, due to Curand's use of the Box-Muller algorithm which produces random values in pairs. Essentially the same as the mumaxplus fix (https://github.com/mumax/plus/pull/155). 

Instead of doing a cuda memcpy, it is done by the existing cuda kernel. In the cuda kernel the size of B_th and noise for the purpose of the kernel uses size of B_th, so since B_th is size N and noise is size N+1, the kernel uses N, and the final float inside noise is essentially invisible and unused. 

In that discussion, there were some concerns over speed and memory usage when using a buffer instead of just writing the values directly. Mumax3 currently uses a buffer. The speed difference was negligible. For a test grid size of ~515x515x55, there is a memory savings of ~ 53 MB for directly writing, or 0.9% of the ~6GB total memory required for the simulation. This seemed small enough that I just kept the buffer method for both N even or odd, since it seems a bit cleaner. The biggest concern is that some future change might accidentally use the N+1 size of noise, but the noise buffer doesn't get touched in the code outside of this section, so the odds of the N+1 size being unintentionally used elsewhere in a future update is small. (I spent a little bit of time trying to see if I could get mumaxplus to avoid needing the buffer for odd N, but it's structured a little differently, seems like it'd be a bit more of a headache to track the hidden float.)

When reviewing, probably the two big things to check is there is no unintentional buffer overflow, and that the N+1 value never gets used unintentionally, but given how noise is used I think I avoided that.

This does mean that for generated values in the stream in N odd simulations, 1 float gets consumed without being used per component, per time step. Otherwise the random value streams line up exactly as before, as expected. As example, I've attached an example script along with a few ovf files for a grid of 3x3x3 and 3x3x2. They match exactly up until the value 0.14437014 (missing from step0 3x3x3. this also corresponds exactly to the bottom of the first column/top of the second column). You see the same thing in future steps, every 3x3x3+1=27+1th value gets skipped, as expected, but otherwise the stream values match exactly.

[oddtest.txt](https://github.com/user-attachments/files/28923918/oddtest.txt)
[Thermal field000000_332.txt](https://github.com/user-attachments/files/28923919/Thermal.field000000_332.txt)
[Thermal field000000_333.txt](https://github.com/user-attachments/files/28923920/Thermal.field000000_333.txt)
[Thermal field000001_332.txt](https://github.com/user-attachments/files/28923921/Thermal.field000001_332.txt)
[Thermal field000001_333.txt](https://github.com/user-attachments/files/28923922/Thermal.field000001_333.txt)

I didn't include a test script, but the example could probably be converted to one fairly easily if it's worth it.

Best regards,
Josh L.

### DIFF
--- a/cuda/curand/generator.go
+++ b/cuda/curand/generator.go
@@ -49,4 +49,11 @@ func (g Generator) SetSeed(seed int64) {
 	}
 }
 
+func (g Generator) SetOffset(offset int64) {
+	err := Status(C.curandSetGeneratorOffset(C.curandGenerator_t(unsafe.Pointer(uintptr(g))), C.ulonglong(offset)))
+	if err != SUCCESS {
+		panic(err)
+	}
+}
+
 // Documentation was taken from the curand headers.
--- a/cuda/temperature.go
+++ b/cuda/temperature.go
@@ -8,6 +8,8 @@ import (
 // Set Bth to thermal noise (Brown).
 // see temperature2.cu
 func SetTemperature(Bth, noise *data.Slice, k2mu0_Mu0VgammaDt float64, Msat, Temp, Alpha MSlice) {
+
+	// N is set by the length of Bth. noise is now set to length N+N%2 (N when N is even, N+1 when N is odd) in engine/temperature.go, so noise[N] is a dummy value that should never be used. The kernel only cares about the underlying pointer and the explicit N count, not the declared shape.
 	util.Argument(Bth.NComp() == 1 && noise.NComp() == 1)
 
 	N := Bth.Len()
--- a/engine/temperature.go
+++ b/engine/temperature.go
@@ -28,8 +28,8 @@ type thermField struct {
 }
 
 func init() {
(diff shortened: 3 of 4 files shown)


JonathanMaes:
Hi Josh,

Thanks for yet another contribution!

I adapted your example to a test script.
For this, I needed to reset the thermal noise within the script, but as we already knew `ThermSeed(seed)` does not yield reproducible noise. I fixed this by adding a new function `ThermReseed(seed)` that not only sets the seed, but also resets the RNG offset back to 0. This seems to be sufficient to fully reset the thermal noise to a reproducible sequence after a call to `ThermReseed`.

The reason for adding a new function, rather than fixing the original `ThermSeed(int)`, is to ensure backwards compatibility. Warnings have been added to mark `ThermSeed` as deprecated. `ThermSeed` and `ThermReseed` behave the same when called before any simulation steps are taken.

Best regards,
Jonathan

PR #410, opened 2026-07-17
CUDA Graph optimization for massive speedup in small systems

JonathanMaes:
*This PR captures part of the changes introduced in [mumax³-CO](https://github.com/mirryou-maker/mumax3-CO), as presented in [C.-Y. You, J. Magn. 31, 204-213 (2026)](https://doi.org/10.4283/JMAG.2026.31.2.204).*

---

As extensively described in [C.-Y. You, J. Magn. 31, 204-213 (2026)](https://doi.org/10.4283/JMAG.2026.31.2.204) and on the [accompanying website](https://sites.google.com/view/spinlab/research/mumax-co), using CUDA Graph to capture and replay the torque-evaluation kernels in each solver stage can result in a speedup of up to 3-5x for small systems, depending on the solver used. This integrates seamlessly into `Run()` and `Steps()` (no changes to input scripts required) and is enabled by default in simulations with <800k cells when >20 steps are requested. Importantly, this optimization is **not used/usable with**

- Nonzero temperatures
- Time-dependent `B_ext` or material parameters
- Custom fields (`AddFieldTerm`)
- `NoDemagSpins`

and is only implemented for the Heun, RK23, RK45DP and RK56 solvers.

### TODO

- [x] Add [C.-Y. You, J. Magn. 31, 204-213 (2026)](https://doi.org/10.4283/JMAG.2026.31.2.204) to the list of suggested references if the optimization is used.
- [x] Add an in-script switch (e.g., `EnableCUDAgraphs`) to disable the usage of CUDA Graphs, in case this causes issues on someone's system.
- [x] Accelerate `minimize()` and `relax()` in the same manner.
- [ ] Test whether everything works as intended on more complex scripts.


engine/graphrun.go
```diff
@@ -0,0 +1,1045 @@
+package engine
+
+// CUDA Graph capture PoC (see CLAUDE.md Phase 1 item 3 / sections 2.7-2.10).
```
JonathanMaes: TODO: remove all references to `CLAUDE.md`, that file does not exist. Refer to the paper instead, if possible.

engine/graphrun.go
```diff
@@ -0,0 +1,1045 @@
...
+	"math"
+
+	"github.com/mumax/3/cuda"
+	"github.com/mumax/3/cuda/cu"
+	"github.com/mumax/3/data"
+	"github.com/mumax/3/util"
+)
+
+func init() {
+	DeclFunc("RunGraph", RunGraph, "Like Steps, but captures the torque-evaluation kernels as CUDA Graphs and "+
```
JonathanMaes: Does this function serve any purpose? It seems that `tryRunGraph` is the version of this function that is actually used. The only use case for this function I see is to print the reason for not using CUDA graphs during a panic. Also, seems like this would more appropriately be called `StepsGraph`.

jplauzie:
Hi,

Would it be possible to extend this it include the minimizer? It has a similar structure, where there are a bunch of consecutive Cuda calls.

(I don't have access to Cuda 13.0, so I'm not sure how the Cuda graph stuff works and whether this is a reasonable ask. If not, feel free to ignore)

engine/graphrun.go
```diff
@@ -0,0 +1,1048 @@
...
+	"github.com/mumax/3/data"
+	"github.com/mumax/3/util"
+)
+
+var EnableCUDAgraphs = true
+
+func init() {
+	DeclFunc("RunGraph", RunGraph, "Like Steps, but captures the torque-evaluation kernels as CUDA Graphs and "+
+		"replays them for the remaining steps. Supports Heun, RK23, RK45DP (default), RK56 and BackwardEuler.")
+	DeclVar("EnableCUDAgraphs", &EnableCUDAgraphs, "Enables CUDA Graphs, greatly improving performance of Run() and Steps() on small grids (default=true)<br>NOTE: graphs are only used if Temp=0, NoDemagSpins=0 and no custom/time-varying fields are defined.")
```
JonathanMaes: How to call this? `EnableCUDAgraphs`? `EnableCUDAgraphCapture`?

engine/graphrun.go
```diff
@@ -0,0 +1,1048 @@
...
+// torqueFn(dy0) and torqueFn(dy) are each captured once, before the loop, as
+// small graphs execT0/execT1 (the kernel sequence computing the torque from
+// M is independent of dt and of M's data values -- only the addresses dy0,
+// dy and M's buffer matter, and those stay fixed for the whole call). The
+// loop body then mirrors Heun.Step exactly, replacing the two torqueFn calls
+// with execT0.Launch/execT1.Launch and running Madd2/Madd3/MaxVecDiff/
+// normalize/adaptDt as ordinary calls on the capture stream (capture has
+// already ended by this point, so the blocking MemCpyDtoH inside MaxVecDiff
+// is allowed).
+func runGraphAdaptive(heun *Heun, condition func() bool) bool {
```
JonathanMaes: `heun` is unused

engine/graphrun.go
```diff
@@ -0,0 +1,1048 @@
...
+	// into the captured graph would be frozen at capture-time Time and
+	// replayed unchanged forever.
+	materialParams := []*regionwise{
+		&Msat.regionwise, &Aex.regionwise, &Dind.regionwise, &Dbulk.regionwise,
+		&Ku1.regionwise, &Ku2.regionwise, &Kc1.regionwise, &Kc2.regionwise, &Kc3.regionwise,
+		&AnisU.regionwise, &AnisC1.regionwise, &AnisC2.regionwise,
+		&B1.regionwise, &B2.regionwise,
+		&Alpha.regionwise, &Xi.regionwise, &Pol.regionwise, &Lambda.regionwise,
+		&EpsilonPrime.regionwise, &FrozenSpins.regionwise, &FreeLayerThickness.regionwise,
+	}
```
JonathanMaes, line 177: TODO: check if this list is complete

JonathanMaes:
Hi Josh,

I had been thinking the same thing, since the use of `minimize()` or `relax()` is probably more common in real scripts than `run()` or `steps()` under the aforementioned conditions. I'll have a look at it once I have checked the existing solvers.

Also, I think CUDA Graphs were introduced in CUDA 10.0, so not having access to 13.0 should not pose an issue. The mention of 13.0 in the other pull request is a wholly separate issue that was already present when trying to compile mumax3.12 with CUDA 13.0.

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+// Type of a graph node, as returned by GraphNode.GetType.
+type GraphNodeType int
+
+const (
+	GRAPH_NODE_TYPE_KERNEL           GraphNodeType = C.CU_GRAPH_NODE_TYPE_KERNEL
+	GRAPH_NODE_TYPE_MEMCPY           GraphNodeType = C.CU_GRAPH_NODE_TYPE_MEMCPY
+	GRAPH_NODE_TYPE_MEMSET           GraphNodeType = C.CU_GRAPH_NODE_TYPE_MEMSET
+	GRAPH_NODE_TYPE_HOST             GraphNodeType = C.CU_GRAPH_NODE_TYPE_HOST
+	GRAPH_NODE_TYPE_GRAPH            GraphNodeType = C.CU_GRAPH_NODE_TYPE_GRAPH
+	GRAPH_NODE_TYPE_EMPTY            GraphNodeType = C.CU_GRAPH_NODE_TYPE_EMPTY
```
JonathanMaes: `GRAPH_NODE_TYPE_`s below this are not supported on all CUDA versions against which mumax³ is compiled. They were introduced in:

- CUDA 11.1: `WAIT_EVENT` and `EVENT_RECORD`
- CUDA 11.2: `EXT_SEMAS_SIGNAL` and `EXT_SEMAS_WAIT`
- CUDA 11.4: `MEM_ALLOC` and `MEM_FREE`
- CUDA 11.7: `BATCH_MEM_OP`
- CUDA 12.3: `CONDITIONAL`

However, these node-type variables don't seem to be used anywhere in the code, so at first sight I think these can be omitted.

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+type GraphNode uintptr
+
+// Controls how StreamBeginCapture interacts with capture sequences on other streams.
+type StreamCaptureMode int
+
+const (
+	STREAM_CAPTURE_MODE_GLOBAL       StreamCaptureMode = C.CU_STREAM_CAPTURE_MODE_GLOBAL
+	STREAM_CAPTURE_MODE_THREAD_LOCAL StreamCaptureMode = C.CU_STREAM_CAPTURE_MODE_THREAD_LOCAL
+	STREAM_CAPTURE_MODE_RELAXED      StreamCaptureMode = C.CU_STREAM_CAPTURE_MODE_RELAXED
+)
```
JonathanMaes, line 50: Stream capture modes were introduced in CUDA 10.1, while mumax³ is built against CUDA &geq;10.0.

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+	BlockDimY      int
+	BlockDimZ      int
+	SharedMemBytes int
+	KernelParams   []unsafe.Pointer
+}
+
+// Begins graph capture on a stream. While capturing, operations enqueued on
+// the stream are not executed but recorded into a graph (see EndCapture).
+func StreamBeginCapture(stream Stream, mode StreamCaptureMode) {
+	err := Result(C.cuStreamBeginCapture(C.CUstream(unsafe.Pointer(uintptr(stream))), C.CUstreamCaptureMode(mode)))
```
JonathanMaes, line 55: `CUstreamCaptureMode()` was introduced in CUDA 10.1, but mumax³ is compiled against CUDA &geq;10.0.

I would simply drop support for 10.0 (released 8y ago). This particular issue can be fixed by supporting &geq;10.1 (released 7.5y ago), but perhaps it would be more convenient to simply support &geq;11.0 (released 6 years ago) at this point.

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+// instantiated executable graph, without re-instantiating it. Used to
+// update per-step scalar arguments (e.g. dt, Time) of a captured kernel
+// launch before replaying the graph with GraphLaunch.
+func GraphExecKernelNodeSetParams(exec GraphExec, node GraphNode, p KernelNodeParams) {
+	cparams, argv, argp := newCKernelNodeParams(&p)
+	if argv != nil {
+		defer C.free(argv)
+		defer C.free(argp)
+	}
+	err := Result(C.cuGraphExecKernelNodeSetParams(C.CUgraphExec(unsafe.Pointer(uintptr(exec))), C.CUgraphNode(unsafe.Pointer(uintptr(node))), &cparams))
```
JonathanMaes: `cuGraphExecKernelNodeSetParams` was introduced in CUDA 10.1, but mumax³ is compiled against CUDA ≥10.0.

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+// the stream are not executed but recorded into a graph (see EndCapture).
+func StreamBeginCapture(stream Stream, mode StreamCaptureMode) {
+	err := Result(C.cuStreamBeginCapture(C.CUstream(unsafe.Pointer(uintptr(stream))), C.CUstreamCaptureMode(mode)))
+	if err != SUCCESS {
+		panic(err)
+	}
+}
+
+// Begins graph capture on this stream.
+func (stream Stream) BeginCapture(mode StreamCaptureMode) {
```
JonathanMaes, line 63: Useless?

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+	var graph C.CUgraph
+	err := Result(C.cuStreamEndCapture(C.CUstream(unsafe.Pointer(uintptr(stream))), &graph))
+	if err != SUCCESS {
+		panic(err)
+	}
+	return Graph(uintptr(unsafe.Pointer(graph)))
+}
+
+// Ends graph capture on this stream, returning the captured graph.
+func (stream Stream) EndCapture() Graph {
```
JonathanMaes, line 78: Useless?

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+	return StreamEndCapture(stream)
+}
+
+// Creates an empty graph.
+func GraphCreate() Graph {
```
JonathanMaes: Useless?

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	nodes := make([]GraphNode, int(n))
+	for i, c := range cnodes[:n] {
+		nodes[i] = GraphNode(uintptr(unsafe.Pointer(c)))
+	}
+	return nodes
+}
+
+// Returns the nodes that make up this graph.
+func (graph Graph) GetNodes() []GraphNode {
```
JonathanMaes: Nowhere used?

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+	var t C.CUgraphNodeType
+	err := Result(C.cuGraphNodeGetType(C.CUgraphNode(unsafe.Pointer(uintptr(node))), &t))
+	if err != SUCCESS {
+		panic(err)
+	}
+	return GraphNodeType(t)
+}
+
+// Returns the type of this graph node.
+func (node GraphNode) GetType() GraphNodeType {
```
JonathanMaes: Nowhere used?

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+		GridDimZ:       int(cparams.gridDimZ),
+		BlockDimX:      int(cparams.blockDimX),
+		BlockDimY:      int(cparams.blockDimY),
+		BlockDimZ:      int(cparams.blockDimZ),
+		SharedMemBytes: int(cparams.sharedMemBytes),
+	}
+}
+
+// Returns the parameters of this kernel-launch graph node. KernelParams is left nil; see KernelNodeParams.
+func (node GraphNode) KernelNodeGetParams() KernelNodeParams {
```
JonathanMaes: Nowhere used?

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+		defer C.free(argp)
+	}
+	err := Result(C.cuGraphKernelNodeSetParams(C.CUgraphNode(unsafe.Pointer(uintptr(node))), &cparams))
+	if err != SUCCESS {
+		panic(err)
+	}
+}
+
+// Sets the parameters of this kernel-launch graph node (the graph must not yet be instantiated).
+func (node GraphNode) KernelNodeSetParams(p KernelNodeParams) {
```
JonathanMaes: Nowhere used?

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+		defer C.free(argp)
+	}
+	err := Result(C.cuGraphExecKernelNodeSetParams(C.CUgraphExec(unsafe.Pointer(uintptr(exec))), C.CUgraphNode(unsafe.Pointer(uintptr(node))), &cparams))
+	if err != SUCCESS {
+		panic(err)
+	}
+}
+
+// Updates the kernel-launch parameters of a node within this executable graph. See GraphExecKernelNodeSetParams.
+func (exec GraphExec) KernelNodeSetParams(node GraphNode, p KernelNodeParams) {
```
JonathanMaes: Nowhere used?

cuda/cu/graph.go
```diff
@@ -0,0 +1,302 @@
...
+func (node GraphNode) KernelNodeSetParams(p KernelNodeParams) {
+	GraphKernelNodeSetParams(node, p)
+}
+
+// Instantiates a graph as an executable graph, ready to be launched
+// repeatedly with GraphLaunch (and updated in place with
+// GraphExecKernelNodeSetParams without re-instantiating).
+func GraphInstantiate(graph Graph) GraphExec {
+	var exec C.CUgraphExec
+	err := Result(C.cuGraphInstantiate(&exec, C.CUgraph(unsafe.Pointer(uintptr(graph))), C.ulonglong(0)))
```
JonathanMaes: This `cuGraphInstantiate` signature is only valid on CUDA &geq;12.0. On CUDA <12.0, the signature `CUresult CUDAAPI cuGraphInstantiate(CUgraphExec *phGraphExec, CUgraph hGraph, CUgraphNode *phErrorNode, char *logBuffer, size_t bufferSize);` was used, which is available as `cuGraphInstantiate_v2` in CUDA &geq;12.0.

Just like the 13.0 bug from #409, this can be fixed by changing the `import "C"` preamble to
```C
#include <cuda.h>
CUresult CUDAAPI cuGraphInstantiate_v2(CUgraphExec *phGraphExec, CUgraph hGraph, CUgraphNode *phErrorNode, char *logBuffer, size_t bufferSize);
```

engine/graphrun.go
```diff
@@ -0,0 +1,1048 @@
...
+}
+
+// runGraphFixedDt implements RunGraph for FixDt != 0: the entire step
+// (StepCaptureBody) is captured once into a single graph and replayed
+// unchanged for n steps.
+func runGraphFixedDt(heun *Heun, condition func() bool) bool {
+	SanityCheck()
+	pause = false
+
+	warmupTorque(M.Buffer())
```
JonathanMaes, line 306: Putting this here without anything else results in an exception 0xc0000005 (Access Violation), as expected since additional buffers will be allocated during `heun.StepCaptureBody()`, which is explicitly disallowed per the `warmupTorque` documentation. The solution is to properly incorporate `StepCaptureBody` into `runGraphFixedDt`.

engine/graphrun.go
```diff
@@ -0,0 +1,1048 @@
...
+			fellBack = runGraphAdaptive(s, condition)
+		}
+	case *RK45DP:
+		fellBack = runGraphRK45DP(s, condition)
+	case *RK23:
+		fellBack = runGraphRK23(s, condition)
+	case *RK56:
+		fellBack = runGraphRK56(s, condition)
+	case *BackwardEuler:
+		fellBack = runGraphBackwardEuler(s, condition)
```
JonathanMaes: Strange ordering. Put these, and the function definitions in the correct `SetSolver()` order. Could even consider moving the function definitions to the separate solver Go files.

engine/graphrun.go
```diff
@@ -0,0 +1,1048 @@
...
+// tryRunGraph) finishes the remaining steps via RunWhile(condition), which
+// re-derives everything from the (possibly changed) global state and is
+// correct regardless of what f changed.
+//
+// Inject is only ever sent to from GUI-connected runs (engine/gui.go) or
+// InjectAndWait (engine/render.go); RunInteractive's keepalive injector only
+// runs after EvalFile returns (cmd/mumax3/main.go). So for headless runs --
+// including all 176 regression tests -- checkInject always returns false and
+// this entire mechanism is a no-op.
+func checkInject() bool {
```
JonathanMaes, line 273: Simply having the GUI open already seems to be sufficient to return true? Need to have a closer look still

engine/graphrun.go
```diff
@@ -0,0 +1,1048 @@
...
+
+		t0 := Time
+		data.Copy(m0, m)
+
+		h := float32(Dt_si * GammaLL)
+
+		// stage 2
+		Time = t0 + (1./5.)*Dt_si
+		cuda.Madd2(m, m, rk.k1, 1, (1./5.)*h)
+		M.normalize()
```
JonathanMaes: I don't see why these `M.normalize()` calls couldn't be included in the graph?

Update: tested for RK56, seems to give the exact same results but with a 3% performance increase, so that's nice.

engine/graphrun.go
```diff
@@ -0,0 +1,1048 @@
...
+//
+// BackwardEuler requires FixDt != 0 (Dt_si = FixDt is unconditional in
+// Step()), so dt is constant for the whole call. The dt-dependent
+// Madd2/normalize calls, the error estimate (MaxVecDiff, which performs the
+// blocking MemCpyDtoH that stream capture cannot record, see CLAUDE.md
+// section 2.8(c)) and NSteps++/setLastErr all run as ordinary (uncaptured)
+// calls in the replay loop, mirroring Step() exactly. The predictor step
+// (Madd2(y, y0, dy1, 1, dt); M.normalize()) is unconditional here because
+// assertGraphCompatible() guarantees Temp.isZero().
+func runGraphBackwardEuler(s *BackwardEuler, condition func() bool) bool {
```
JonathanMaes: Backward Euler is not a real backward Euler in the original implementation, so I would not graph-accelerate this until the backward Euler has been updated to be a true backward Euler method.

JonathanMaes (commented):
WIP

Issue #329, opened 2023-11-18
Minimizer() gives inconsistent results for the same input, between runs and in the same run. 


jsampaio:
The Minimize() function runs for a different number of steps for the same input state, and gives a different end state.

Demonstrator code below.
In this code, minimize() is called five times with exactly the same initial magnetisation m (in a for loop), and the output is checked (exemplified here by the magnetisation at an arbitrary cell). This script was run several times.
**Result:** The end result is different everytime the script is run, as well as for every one of the executions with the same initial state (in the for loop).
**Expected Result:** the same input should produce exactly the same output. Specially, between runs of the script, but also between runs with the same initial state.

This is a problem because:

- the user expects the same file to produce the same output.
- Often the user will compare different responses of what he expects is the same initial state. If the initial state is produced with Minimize(), the initial state will be different at every run, introducing unexpected differences.

I've quickly checked the code of engine/minimizer.go and cuda/minimize.cu and it is not clear to me what causes this random behaviour.

### Demonstrator code:
```
SetMesh(128, 64, 1, 4e-9, 4e-9, 4e-9, 0, 0, 0); 
Msat = 1e6; Aex = 10e-12; alpha = 1.;
tableaddvar(step, "step", "");
tableadd(MaxTorque);
tableAdd(Crop(m, 30, 31, 30, 31, 0, 1)); // mag in one arbitrary cell (more sensitive to small differences than the whole average)

for i := 0; i < 5; i++ {
	m = TwoDomain(0.,0,1., 0,1,0, 0,0.,-1.);
	step=0;
	Minimize(); 
	tablesave(); print(t, step, MaxTorque, Crop(m, 30, 31, 30, 31, 0, 1).average()); 
}
```

### Outputs.
Here's the output for two runs on the same computer. The columns are:

- time (always 0)
- The number of steps that Minimize() used
- The final MaxTorque
- The m at an arbitrary cell (30,30,1).

Run 1:
```
//0 1708 1.3587993969904421e-06 [0.24186624586582184 -0.9649359583854675 0.10197720676660538]
//0 706 1.9388212427259442e-06 [0.7151626944541931 -0.6850084662437439 -0.13894522190093994]
//0 1622 1.4579250321644758e-06 [-0.7151928544044495 0.6849732398986816 -0.13896377384662628]
//0 1650 1.943069133851686e-06 [0.5473531484603882 -0.8349050283432007 0.057777922600507736]
[...]

PR #415, opened 2026-08-28
Make reduction operations deterministic

JonathanMaes:
As reported in #329 and several times on the mailing list, `minimize()` could give different results on the same hardware when running the exact same script multiple times.

The reason for this, it turns out, is the nonassociativity of floating-point calculation.
The minimizer uses `cuda.Dot`, which is a reduction operation. In mumax³, reductions are calculated on the GPU in 8 parallel blocks, the results of each of which are then added together to obtain the final result. However, depending on which block finishes first, these values can be added in a different order, potentially yielding different results due to the aforementioned nonassociativity. 

This also affected `cuda.Sum`, which for example caused slight variations in `E_anis`, as demonstrated by this small example script:

```go
SetCellSize(4e-9, 4e-9, 4e-9)
SetGridSize(128, 128, 2)
Msat = 800e3
Ku1 = 1e5
AnisU = Vector(1, 0, 0)
m = RandomMagSeed(2)
print(E_anis)
```

This effect does not seem to be present in the other reduction operations of `cuda/reduce.go`, since they all return a maximum value and are therefore not affected by the nonassociativity.

### Solution

A new macro `reduceDeterministic` can now is now used by the affected CUDA kernels, that is largely similar to the existing `reduce` except that it populates an array of values (with length = number of blocks) rather than a single scalar value. On the Go side, the relevant functions then simply add the elements of this array in their order of appearance, yielding a deterministic value due to the indexing used. It seems that this has a negligible performance impact.

### Possible improvements

The aggregate values of each block could also be sorted in ascending order, to perform the most accurate floating-point addition possible. However, this would incur additional overhead, and would in practice not actually contribute to bringing results across different GPUs closer together, so I don't think sorting is useful.

An alternative solution could be to [specify the determinism level](https://developer.nvidia.com/blog/controlling-floating-point-determinism-in-nvidia-cccl), as suggested by @jplauzie elsewhere, but this seems to have a more noticeable performance impact.

Also, instead of 8 blocks, the number of blocks could be increased to equal the number of multiprocessors on the GPU. I believe the currently used value of 8 might just have been a typical number for the GPUs at the time mumax³ was written. I am not sure how much this would affect performance and which amount of blocks would be ideal.
> Update: it seems that using an equal amount of blocks as there are SM's on the GPU, as was ultimately implemented in this PR, results in a >2% speedup of `minimize()` overall (for a 2048x2048 simulation; smaller systems experience a smaller performance increase).

### DIFF
--- a/cuda/init.go
+++ b/cuda/init.go
@@ -47,6 +47,8 @@ func Init(gpu int) {
 		log.Println("DEBUG: synchronized CUDA calls")
 	}
 
+	reducecfg = &config{Grid: cu.Dim3{X: cu.DeviceGetAttribute(cu.MULTIPROCESSOR_COUNT, dev), Y: 1, Z: 1}, Block: cu.Dim3{X: REDUCE_BLOCKSIZE, Y: 1, Z: 1}}
+
 	// test PTX load so that we can catch CUDA_ERROR_NO_BINARY_FOR_GPU early
 	fatbinLoad(madd2_map, "madd2")
 }
--- a/cuda/reduce.go
+++ b/cuda/reduce.go
@@ -18,9 +18,16 @@ const REDUCE_BLOCKSIZE = C.REDUCE_BLOCKSIZE
 // Sum of all elements.
 func Sum(in *data.Slice) float32 {
 	util.Argument(in.NComp() == 1)
-	out := reduceBuf(0)
-	k_reducesum_async(in.DevPtr(0), out, 0, in.Len(), reducecfg)
-	return copyback(out)
+	partials := reduceBufN(reducecfg.nBlocks(), 0)
+	k_reducesum_async(in.DevPtr(0), partials, 0, in.Len(), reducecfg)
+	host := copybackSlice(partials, reducecfg.nBlocks())
+
+	// Add elements of partials on CPU
+	var result float32
+	for _, v := range host {
+		result += v
+	}
+	return result
 }
 
 // Dot product.
--- a/cuda/reduce.go
+++ b/cuda/reduce.go
@@ -137,6 +137,5 @@ func initReduceBuf() {
 }
 
 // launch configuration for reduce kernels
-// 8 is typ. number of multiprocessors.
-// could be improved but takes hardly ~1% of execution time
-var reducecfg = &config{Grid: cu.Dim3{X: 8, Y: 1, Z: 1}, Block: cu.Dim3{X: REDUCE_BLOCKSIZE, Y: 1, Z: 1}}
+// Grid size is set during cuda.Init to equal the number of multiprocessors.
+var reducecfg *config
--- a/cuda/reduce.go
+++ b/cuda/reduce.go
@@ -27,12 +27,19 @@ func Sum(in *data.Slice) float32 {
 func Dot(a, b *data.Slice) float32 {
 	nComp := a.NComp()
 	util.Argument(nComp == b.NComp())
-	out := reduceBuf(0)
+	partials := reduceBufN(reducecfg.nBlocks(), 0)
 	// not async over components
 	for c := 0; c < nComp; c++ {
-		k_reducedot_async(a.DevPtr(c), b.DevPtr(c), out, 0, a.Len(), reducecfg) // all components add to out
+		k_reducedot_async(a.DevPtr(c), b.DevPtr(c), partials, 0, a.Len(), reducecfg) // all components add to out
 	}
-	return copyback(out)
+	host := copybackSlice(partials, reducecfg.nBlocks())
+
+	// Add elements of partials on CPU
+	var result float32
+	for _, v := range host {
+		result += v
+	}
+	return result
 }
 
 // Maximum of absolute values of all elements.
@@ -65,7 +72,8 @@ func MaxVecDiff(x, y *data.Slice) float64 {
 	return math.Sqrt(float64(copyback(out)))
 }
 
-var reduceBuffers chan unsafe.Pointer // pool of 1-float CUDA buffers for reduce
+var reduceBuffers chan unsafe.Pointer                // pool of 1-float CUDA buffers for reduce
+var reduceBuffersN = map[int64]chan unsafe.Pointer{} // pool of N-float CUDA buffers for reduce
 
 // return a 1-float CUDA reduction buffer from a pool
 // initialized to initVal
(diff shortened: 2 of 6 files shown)


jplauzie:
A very nice solution!

>Also, instead of 8 blocks, the number of blocks could be increased to equal the number of multiprocessors on the GPU. I believe the currently used value of 8 might just have been a typical number for the GPUs at the time mumax³ was written. I am not sure how much this would affect performance and which amount of blocks would be ideal.

I am not a Cuda expert, but 8 is likely to be an underutilization. Nvidia recommends at least as many blocks as SMs, and preferably at least a few blocks per SM is best (so it can issue instructions for one block while waiting for another). The number of blocks can be quite large, potentially hundreds. 

Mumax3 uses 512 threads/block (mumaxplus is 256. Both are fine), so 8 blocks would lock you to only 512*8=4,096 threads. The maximum is ~2048 threads per SM, so that would be two SMs worth. For context, my 1070ti has 19 SMs, and newer cards will have more ( ~150ish). However, there are other limiters, like shared memory usage, register pressure, warp schedulers etc. For reduction kernels because they are very light computation, Nvidia recommends to look at memory bandwidth, and try to maximize that. That said, from what I can tell, the exact value doesn't really matter that much, as long as you are creating enough work to keep the GPU busy and hide latency.

To give some idea, running the reduction kernel k_transRotSumsPartial i wrote for [mumaxplus](https://github.com/mumax/plus/pull/171) (it is not quite the same, because of the weighting and it is doing two reductions at once, but it is very very close to a normal sum reduction kernel), I get something like (this is for a very large grid, 1024x1024x1):

blocksize, timing, memory bandwidth usage
1 block, 2.232 ms per kernel call, 5.64 GB/s
8 blocks, .3 ms,  41 GB/s
64 blocks, 0.077 ms, 164.25 GB/s
152 blocks 0.071ms, 170 GB/s

for a small grid (128x32x1), 8 and 16 about the ~same, and 1 block is half as slow.  (152 blocks at 256 threads =38912 , or 38912/19=2048 threads/SM). The theoretical peak memory bandwidth for my GPU is 256 GB/s, so ~66%. I am still trying to see if i can get it up more, but I don't know any other tricks, and I don't know yet if mine is a bit lower because of the extra ops. Mark Harris' [guide](https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf) for a sum reduction is able to get up to something like 85% bandwidth (73/86.4 GB/s), which is a good value to shoot for. To get the 152 number, I used something like:

```
constexpr int REDUCTION_OCCUPANCY_MULTIPLIER = 1;
template <typename Kernel>
int occupancyMaxBlocks(Kernel kernel, int blockDim, size_t sharedMemBytes = 0) {
  static int cached = -1;
  if (cached < 0) {
    int device;
    checkCudaError(cudaGetDevice(&device));
    int smCount;
    checkCudaError(cudaDeviceGetAttribute(&smCount, cudaDevAttrMultiProcessorCount, device));

    int maxActiveBlocksPerSm = 0;
    checkCudaError(cudaOccupancyMaxActiveBlocksPerMultiprocessor(
        &maxActiveBlocksPerSm, kernel, blockDim, sharedMemBytes));

    cached = smCount * maxActiveBlocksPerSm * REDUCTION_OCCUPANCY_MULTIPLIER;
  }
  return cached;
}
```

The Cuda API lets you query for things like number of SMs, and cudaOccupancyMaxActiveBlocksPerMultiprocessor does a little bit of heuristics to try to find a rough estimate to the correct number of blocks per thread (it even seems to look at things like shared memory constraints, not just threads, which is very nice). For the parallel reduction tree setup, you'd want to round it to a power of 2 presumably.

It is only occupancy though, so sometimes you can maybe beat that number via handtuning. You can have lower occupancy but better performance as long as you can keep things busy. But it is a good place to start. (The Cuda toolkit also used to have a spreadsheet somewhere that is supposed to help with this. Deprecated in favor of Nsight Compute).

jplauzie:
Taking another look, the 8 number probably comes from Mark Harris's presentation (the one linked above is the same one that Arne's comment refers to). Slide 5 shows 8 blocks. But even later on, slide 31 it says "best perf with 64-256 blocks of 128 threads", so I think 8 was just for example, to visually show the tree reduction. Makes sense based on the logic above, a G80 had 16 SMs with 8 blocks per SM, which gets you the 64-256 blocks). You can also find examples (see e.g. : https://developer.nvidia.com/blog/faster-parallel-reductions-kepler/ and the github it refers to) with much larger choices of blocks, even back in ~2014.

There was another comment from Arne that the reduction wasn't optimized, but it was a small part of the run time (which is true, for dynamics. The FFTs are >90% of typical run times, even on my 1070. But not for minimize/hysteresis)

JonathanMaes:
Thanks for the interesting information as always, Josh.

Since the reduction operation isn't performance-critical in a simulation, I think it should suffice to make it as performant as reasonably possible without adding a lot of extra code.

I profiled `reducesum` for a 2048x2048 simulation on a GPU with 48 SM's:

| Blocks in grid | 1 | 8 | 48=SM | 128 | 192 | 256 | 384 | 2048 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Duration [µs] | 969.66 | 127.49 | 44.96 | 44.1 | 45.28 | 59.14 | 44.83 | 48.86 |
| Compute throughput [%] | 0.51 | 3.92 | 11.99 | 13.63 | 14.37 | 15.55 | 17.63 | 42.44 |
| Memory throughput [%] | 3.96 | 29.99 | 85.18 | 86.74 | 84.42 | 64.76 | 85.28 | 78.62 |

Indeed, the previous value of 8 is 3x slower than can be achieved with larger launch grids. Conversely, too many blocks (2048) is probably counterproductive for small simulations (and would move the performance bottleneck to the CPU-side addition of the per-block values, which is not captured in this profiling).

At 192 or higher, Nsight warned me about the "tail effect" affecting performance: "this kernel launch results in 1 full waves and a partial wave of 48 thread blocks" (with those numbers varying with the launch grid size). For 256, this effect seems to have been particularly bad given the drop in performance. The `cudaOccupancyMaxActiveBlocksPerMultiprocessor` you mentioned would probably avoid this by avoiding a too large launch grid, but that function takes the kernel itself as an argument, which I don't think we have easy access to in the Go layer.

Since the duration seems to plateau as soon as the number of blocks equals the number of SM's, I think it should be fine to just use that value, as I have done in the following commit. As always, any concerns regarding this approach are welcome!

jplauzie:
That seems reasonable to me, and that performance seems excellent. ~85% memory bandwidth is exactly what is expected for a peak reduction performance, well done!

I'm not sure exactly why 192 was so drastic (I would guess something to do with 48 SMs, 3 blocks per SM at 512 threads/block gets you to 1,536 threads per SM which is the thread cap per SM for your card. so caps at 48*3=144 blocks?), but capping it at numblocks= SMs seems safe and performant without needing to worry about the details of that.

>Conversely, too many blocks (2048) is probably counterproductive for small simulations

Did 48 seem ok for smaller simulations, as well? For smaller simulations, if it's a concern, perhaps you can cap it lower with some heuristic, if necessary. But if small simulations were ok with numblocks=SMs, this seems like a good solution to leave it at

JonathanMaes:
Indeed, for a 64x64 system, 48 seems to have a similar performance than 8 and 144.
For small systems, all kernel launches take a similar amount of time (3-4µs) as they seem to be overhead-dominated.
Changing the number of blocks does (seemingly linearly) increase the compute and memory throughput percentage, but this is not reflected in an increased runtime of the kernel.

